PR #3, opened 2019-11-27
[WIP] Develop   concept

donhausk:
This is a first version. It should compile and also runs with Turing. Please try to read the code and propose test which need to be made.

The structure is taken from the container file in the Turing package. The idea is to create an own Particle Filter package which allows more freedom. 

Remark: The interface is not at all complete and should be completely rewritten. The only purpose of this interface is for debugging and to demonstrate what the package could be used for.
The main purpose so far is to have a clean separation between particle filter and Turing. The focus is still on making the package independent on Turing. 

Remark: This is a WIP pull and it is clear that this pull will not be accepted.

Many thanks for the comments so far!


Tests/demonstarte.jl
```diff
@@ -0,0 +1,106 @@
...
+
+        #Proposal and Resampling
+        logpy = logpdf(Normal(get_x(var,i),0.4),y[i-1])
+
+        var = report_observation(var,logpy)
+        if var.taskinfo.hold
+            produce(0.0)
+
+    end
+end
```
devmotion: Is this the way how a user could/should use the package without Turing? It seems the user has to reimplement almost the whole algorithm, is that the intention?

src/AdvancedPS.jl
```diff
@@ -0,0 +1,66 @@
+### Attention this is a development package! It wount run.
+
+
+module AdvancedPS
+
+
+
+        const DEBUG = Bool(parse(Int, get(ENV, "DEBUG_APS", "0")))
```
devmotion: Again this should not be indented.

src/Interface.jl
```diff
@@ -0,0 +1,46 @@
...
+# A very light weight container for a state space model
+# The state space is one dimensional!
+# Note that this is only for a very simple demonstration.
+mutable struct Container
+    x::Vector{Float64}
+    num_produce::Float64
+end
+
+# This is important for initalizaiton
+function init()
```
devmotion: I guess you could just write `const init = current_trace`?

src/Interface.jl
```diff
@@ -0,0 +1,46 @@
...
+    return current_trace()
+end
+
+# Th
+function report_observation(trace, logp::Float64)
```
devmotion: Maybe rename the function to something that's more self-explanatory? Additionally, I guess it would be nice to stick to the convention of adding an exclamation mark at the end of function names if functions modify the arguments in-place. I'm not sure if specifying the type of `logp` is useful here, since you're not dispatching on the type you could just remove it.

src/Interface.jl
```diff
@@ -0,0 +1,46 @@
...
+
+function get_x(trace,indx)
+    return @inbounds trace.vi.x[indx]
+end
+function set_x(trace,indx,val)
+    return @inbounds trace.vi.x[indx] = val
+end
+
+
+function copyC(vi::Container)
```
devmotion: Maybe you could just overload `deepcopy`?

src/resample_functions.jl
```diff
@@ -0,0 +1,288 @@
...
+    return rand(Distributions.sampler(Categorical(w)), num_particles)
+
+end
+
+
+
+function resample_residual(w::AbstractVector{<:Real}, num_particles::Integer)
+
+
+
```
devmotion: I guess it would be nice to remove some of the white space?

src/samplers.jl
```diff
@@ -0,0 +1,151 @@
+
+
+function sampleSMC!(pc::ParticleContainer,resampler::Function =resample_systematic ,resampler_threshold::AbstractFloat = 0.5)
```
devmotion: As mentioned in https://github.com/TuringLang/Turing.jl/pull/990, I think one should not implement the resampling based on ESS explicitly. It feels more natural to me to instead implement a resampler that allows to combine basically any resample algorithm with the ESS checks, as I tried to do in https://github.com/TuringLang/Turing.jl/pull/990. The changes in that PR even allow to reuse the weights computed for evaluation the ESS in the resampling step.

src/samplers.jl
```diff
@@ -0,0 +1,151 @@
...
+            n = length(pc)
+            nresamples = n
+            indx = resampler(Ws, nresamples)
+            resample!(pc, indx)
+        end
+    end
+end
+
+# The resampler threshold is only imprtant for the first step!
+function samplePG!(pc::ParticleContainer,resampler::Function = resample_systematic ,ref_particle::Union{Particle,Nothing}=nothing , resampler_threshold =0.5)
```
devmotion: Here (and above in `sampleSMC!`) one shouldn't restrict `resampler` to `Function`s IMO. It's common to implement callable structs (especially for resamplers that might be interesting since it would allow to define cache arrays), which could not be used here.

src/taskinfo.jl
```diff
@@ -0,0 +1,46 @@
...
+    logp::Float64
+
+    # This corresponds to p(x_t|x_t-1)*p(x_t-1|x_t-2)*... *p(x_0)
+    # or |x_{0:t-1} for non markovian models, we need this to compute
+    # the ancestor weights.
+
+    logpseq::Float64
+end
+
+function copy(info::PGTaskInfo)
```
devmotion: That's just a matter of taste, but usually I think it's easier to spot overloaded methods by not importing them explicitly but writing
```suggestion
function Base.copy(info::PGTaskInfo)
```
instead.

src/taskinfo.jl
```diff
@@ -0,0 +1,46 @@
...
+    logpseq::Float64
+
+
+    hold::Bool
+end
+function PGASTaskInfo(logp::Float64,logpseq::Float64)
+    PGASTaskInfo(logp,logpseq,false)
+end
+
+function copy(info::PGASTaskInfo)
```
devmotion: Same here:
```suggestion
function Base.copy(info::PGASTaskInfo)
```

src/Interface.jl
devmotion:
```suggestion
function Base.deepcopy(vi::Container)
```

src/AdvancedPS.jl
```diff
@@ -0,0 +1,66 @@
...
+                resample_multinomial,
+                resample_residual,
+                resample_stratified,
+                resample_systematic
+
+        include("Interface.jl")
+
+        export  report_transition,
+                report_observation,
+                init,
```
devmotion: `init` seems a bit too generic to be exported. Maybe rename it?

src/AdvancedPS.jl
```diff
@@ -0,0 +1,66 @@
...
+
+
+module AdvancedPS
+
+
+
+        const DEBUG = Bool(parse(Int, get(ENV, "DEBUG_APS", "0")))
+        using Libtask
+        using StatsFuns: logsumexp, softmax!
+        import Base.copy
```
devmotion: Not needed (see comments below)
```suggestion
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+"""
+Data structure for particle filters
+- effectiveSampleSize(pc :: ParticleContainer)
+- normalise!(pc::ParticleContainer)
+- consume(pc::ParticleContainer): return incremental likelihood
+"""
+mutable struct ParticleContainer{Tvi,TInfo} <: AbstractParticleContainer where {Tvi, TInfo <: AbstractTaskInfo}
+    vals::Vector{Trace{Tvi,TInfo}}
+    # A named tuple with functions to manipulate the particles vi.
+    manipulators::Dict{String,Function}
```
devmotion: That's not a named tuple but a dictionary? I guess you could just add additional fields for whatever information you want to store? Moreover, as mentioned above, it might not be a good idea to restrict this to `Function`s only.

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+const Particle = Trace
+
+"""
+Data structure for particle filters
+- effectiveSampleSize(pc :: ParticleContainer)
+- normalise!(pc::ParticleContainer)
+- consume(pc::ParticleContainer): return incremental likelihood
+"""
+mutable struct ParticleContainer{Tvi,TInfo} <: AbstractParticleContainer where {Tvi, TInfo <: AbstractTaskInfo}
+    vals::Vector{Trace{Tvi,TInfo}}
```
devmotion: Is it necessary to be so explicit with the types? Couldn't you just define
```julia
mutable struct ParticleContainer{T<:Trace} <: AbstractParticleContainer
    vals::Vector{T}
```
And I don't see immediately why the additional type parameters would be useful?

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+
+end
+
+## Empty initilaizaiton
+ParticleContainer{Tvi,TInfo}() where {Tvi, TInfo <: AbstractTaskInfo} = ParticleContainer{Tvi,TInfo}(Dict{String,Function}(),0)
+ParticleContainer{Tvi,TInfo}(manipulators::Dict{String,Function}) where {Tvi, TInfo <: AbstractTaskInfo} = ParticleContainer{Tvi,TInfo}(manipulators,0)
+#Some initilaizaiton...
+function ParticleContainer{Tvi,TInfo}(manipulators::Dict{String,Function}, n::Int) where {Tvi, TInfo <: AbstractTaskInfo}
+    if !("copy" in keys(manipulators)) manipulators["copy"] = deepcopy end
+    if !("set_retained_vns_del_by_spl!" in keys(manipulators)) manipulators["set_retained_vns_del_by_spl!"] = (x) -> () end
```
devmotion: This doesn't seem right. I guess you should just add fields with default values?

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+    ParticleContainer{Tvi,TInfo}( Vector{Trace{Tvi,TInfo}}(undef, n), manipulators, Float64[],Float64[], 0.0, 0)
```
devmotion: That leads to arrays of different sizes, is that intended?

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+
+end
```
devmotion: In the current implementation in Turing there's no need for any special constructors, in particular not for constructors without particles. All particles could be created first, and then a container could be constructed with them, filling in zeros and other default values for the remaining fields. Did that change?

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+
+
+Base.collect(pc :: ParticleContainer) = pc.vals # prev: Dict, now: Array
+Base.length(pc :: ParticleContainer)  = length(pc.vals)
+Base.similar(pc :: ParticleContainer{T}) where T = ParticleContainer{T}(0)
+# pc[i] returns the i'th particle
+Base.getindex(pc :: ParticleContainer, i :: Real) = pc.vals[i]
```
devmotion: The updated implementation in Turing allows to forward `@inbounds`. Also `Real` is not really useful here, I guess.
```suggestion
Base.@propagate_inbounds Base.getindex(pc::ParticleContainer, i::Int) = pc.vals[i]
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+    if !("copy" in keys(manipulators)) manipulators["copy"] = deepcopy end
+    if !("set_retained_vns_del_by_spl!" in keys(manipulators)) manipulators["set_retained_vns_del_by_spl!"] = (x) -> () end
+    ParticleContainer{Tvi,TInfo}( Vector{Trace{Tvi,TInfo}}(undef, n), manipulators, Float64[],Float64[], 0.0, 0)
+
+end
+
+
+Base.collect(pc :: ParticleContainer) = pc.vals # prev: Dict, now: Array
+Base.length(pc :: ParticleContainer)  = length(pc.vals)
+Base.similar(pc :: ParticleContainer{T}) where T = ParticleContainer{T}(0)
```
devmotion: That's not really a proper implementation of `similar`, since it doesn't respect the size of the current container. It was not needed before, so I just removed it in Turing.
```suggestion
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+# pc[i] returns the i'th particle
+Base.getindex(pc :: ParticleContainer, i :: Real) = pc.vals[i]
+
+
+
+# registers a new x-particle in the container
+function Base.push!(pc::ParticleContainer, p::Particle{Tvi,TInfo}) where {Tvi,TInfo <: AbstractTaskInfo}
```
devmotion: The static parameters are not useful.
```suggestion
function Base.push!(pc::ParticleContainer, p::Particle)
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+    push!(pc.vals, p)
+    push!(pc.logWs, 0.0)
+    push!(pc.logpseq,0.0)
+    pc
+end
+
+function extend!(pc::ParticleContainer, n::Int, varInfo, tasks::Task, taskInfo::T) where T <: AbstractTaskInfo
```
devmotion: In Turing, there's no need for having this specialized `append!` function. Did that change? In any case, better would be
```suggestion
function extend!(pc::ParticleContainer, n::Int, varInfo, tasks::Task, taskInfo::AbstractTaskInfo)
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+    z1 = logZ(pc)
+    n = length(pc)
+
+    particles = collect(pc)
+    num_done = 0
+    for i=1:n
+        p = particles[i]
+        score = Libtask.consume(p)
+
+        if in(:logpseq,fieldnames(typeof(p.taskinfo))) && isa(p.taskinfo.logpseq,Float64)
```
devmotion: Maybe just
```suggestion
        if hasfield(typeof(p.taskinfo), logpseq)
```
or
```suggestion
        if hasproperty(typeof(p.taskinfo), logpseq)
```
? That would require Compat 2.0.0 for support of Julia 1.0 and Julia 1.1.

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+            set_logpseg(pc,i,p.taskinfo.logpseq)
```
devmotion: I guess this should be
```suggestion
            set_logpseq!(pc, i, p.taskinfo.logpseq)
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+    Ws = weights(pc)
+    return inv(sum(abs2, Ws))
+end
+
+increase_logweight(pc :: ParticleContainer, t :: Int, logw :: Float64) =
+    (pc.logWs[t]  += logw)
+
+
+
+set_logpseg(pc :: ParticleContainer, t :: Int, logp :: Float64) =
```
devmotion: I'd suggest
```suggestion
set_logpseq!(pc::ParticleContainer, t::Int, logp::Float64) =
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+        pc.n_consume += 1
+        # res = increase_loglikelihood(pc, z2 - z1)
+    end
+
+    res
+end
+
+
+# compute the normalized weights
+weights(pc::ParticleContainer) = softmax!(copy(pc.logWs))
```
devmotion: There's no need to copy first and then use the in-place version? IMO, more natural would be
```suggestion
weights(pc::ParticleContainer) = softmax(pc.logWs)
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+# compute the log-likelihood estimate, ignoring constant term ``- \log num_particles``
+logZ(pc::ParticleContainer) = logsumexp(pc.logWs)
+
+# compute the effective sample size ``1 / ∑ wᵢ²``, where ``wᵢ```are the normalized weights
+function effectiveSampleSize(pc :: ParticleContainer)
+    Ws = weights(pc)
+    return inv(sum(abs2, Ws))
+end
+
+increase_logweight(pc :: ParticleContainer, t :: Int, logw :: Float64) =
```
devmotion: I'd recommend sticking to the conventions and writing
```suggestion
increase_logweight!(pc::ParticleContainer, t::Int, logw::Float64) =
```

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+    (pc.logWs[t]  += logw)
+
+
+
+set_logpseg(pc :: ParticleContainer, t :: Int, logp :: Float64) =
+    (pc.logpseq[t] = logp)
+
+increase_logevidence(pc :: ParticleContainer, logw :: Float64) =
```
devmotion: Same here, I would recommend
```suggestion
increase_logevidence!(pc::ParticleContainer, logw::Float64) =
```

Tests/demonstarte.jl
donhausk: No, unfortunately, I was not very clear in what the purpose of this pr is. This is just for debugging. A proper interface is certainly something which needs to be implemented. The only purpose of this interface is to demonstrate what the flexibility is.

src/resample_functions.jl
donhausk: Yes, I am sorry. I have copied most of it from Turing. I will remove them!

src/Interface.jl
donhausk: I am sorry, i did not mention that the interface is only for debugging and demonstration. Maybe i should remove it completely out of this version.

src/Interface.jl
donhausk: I am sorry, i did not mention that the interface is only for debugging and demonstration. Maybe i should remove it completely out of this version.

src/Interface.jl
donhausk: I am sorry, i did not mention that the interface is only for debugging and demonstration. Maybe i should remove it completely out of this version.

src/samplers.jl
donhausk: Yes, good point! The focus of this pull is mainly on separating the particle filter form Turing. Your changes are not yet included and should definitely be included.

src/samplers.jl
donhausk: Thanks for that comment!

src/AdvancedPS.jl
donhausk: Good point, I will most probably remove the whole Interface.

src/tasks.jl
donhausk: Good point. I had a discussion with @torfjelde and we are going to make a struct out of it. This is type stable and more flexible.

src/tasks.jl
donhausk: Good point, however, this implies that we need to define the Trace in Turing. One of the goals is that we can make changes in this package without the need for making corrections in Turing.

src/tasks.jl
donhausk: Yes, this is clearly wrong. After a discussin with @torfjelde, we decided to change this quite a lot...

src/tasks.jl
donhausk: This was copied from the original code in the container file....

src/tasks.jl
donhausk: This is actually used in the resample.jl file

src/Interface.jl
devmotion: No worries, just ignore all my comments regarding the debugging functionality :slightly_smiling_face:

src/tasks.jl
devmotion: Can you explain why that would imply that you have to define `Trace` in Turing? I don't see immediately why that would be the case.

src/tasks.jl
devmotion: You mean in https://github.com/TuringLang/AdvancedPS.jl/blob/f1f98652324d527c5ab149c6a9d3fbb319fd03f6/src/resample.jl#L18 ? To me it seems there `particles` is actually of type `Vector{<:Trace}` and not of type `ParticleContainer`, so that doesn't call the method we're discussing here. Removing it in Turing also didn't break anything :slightly_smiling_face:

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+            set_logpseg(pc,i,p.taskinfo.logpseq)
+        end
+
+        if score isa Real
+            score += p.taskinfo.logp
+
+            ## Equivalent to reset logp
+            p.taskinfo.logp = 0
+
+            increase_logweight(pc, i, Float64(score))
```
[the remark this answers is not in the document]
devmotion: Actually, the Julia documentation seems to take a different stance on this: https://docs.julialang.org/en/v1/manual/style-guide/#Handle-excess-argument-diversity-in-the-caller-1

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+
+
+# step to the next observe statement, return log likelihood
+Libtask.consume(t::Trace) = (t.vi.num_produce += 1; consume(t.task))
+
+# Task copying version of fork for Trace.
+function fork(trace :: Trace, copy_vi::Function, is_ref :: Bool = false, set_retained_vns_del_by_spl!::Union{Function,Nothing} = nothing)
+    newtrace = copy(trace,copy_vi)
+    if is_ref
+        @assert set_retained_vns_del_by_spl! != nothing "[AdvancedPF] set_retained_vns_del_by_spl! is not set."
```
devmotion: Another thing: It can be an advantage to compare singletons (such as `missing`, `nothing`, ...) with `===` and `!==` (see, e.g., https://github.com/JuliaLang/julia/issues/28695). It's more important for iterations and cases such as `findfirst` etc., I guess, but I tend to use it everywhere consistently since there's no harm in using it.

src/tasks.jl
```diff
@@ -0,0 +1,233 @@
...
+
+# TaskInfo stores additional informatino about the task
+mutable struct Trace{Tvi, TInfo}  <: AbstractTrace where {Tvi, TInfo <: AbstractTaskInfo}
+    vi::Tvi # Unfortunatley, we can not set force this to be a subtype of VarInfo...
+    task::Task
+    taskinfo::TInfo
+end
+
+function Base.copy(trace::Trace{Tvi,TInfo}, copy_vi::Function) where {Tvi, TInfo <: AbstractTaskInfo}
+    res = Trace{typeof(trace.vi),typeof(trace.taskinfo)}(copy_vi(trace.vi), copy(trace.task), copy(trace.taskinfo))
```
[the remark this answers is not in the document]
devmotion: I guess it's just copied from Turing, there `Trace` is defined with some inner constructors based for `SamplePrior` and regular samplers (not sure if the functionality that one can specify a `SamplePrior` type parameter with a regular sampler as argument is ever used).

src/samplers.jl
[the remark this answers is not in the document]
cpfiffer: I'm not fond of having functions defined with the sampler name in the signature. Turing and most of the packages under it's umbrella use multiple dispatch on some kind of sampler object -- I think good practice would be to do the same here so that everything is consistent across the Turingverse. 

Check out [`AbstractMCMC`](https://github.com/TuringLang/AbstractMCMC.jl/blob/master/src/AbstractMCMC.jl) and the interface methods it provides for a sampler. There's also a fairly detailed writeup about the interface methodology [here](https://github.com/TuringLang/Turing.jl/blob/csp/interface-guide/docs/src/using-turing/interface.md) and a slightly less detailed one [here](https://github.com/TuringLang/Turing.jl/issues/895#issuecomment-530942507).

In this case, I think `pc`, `resampler`, and `resampler_threshold` could be bundled into one `SMCSampler` struct or something.

src/AdvancedPS.jl
```diff
@@ -0,0 +1,66 @@
...
+
+        abstract type AbstractTaskInfo end
+        abstract type AbstractParticleContainer end
+        abstract type AbstractTrace end
+
+
+        include("tasks.jl")
+
+        export  ParticleContainer,
```
cpfiffer: All of these export blocks should be consolidated into one. It's a little hard to track what's being exported where.

src/tasks.jl
devmotion: I assume your point is that Turing needs  a slightly different `Trace`? Then I would suggest
```julia
mutable struct ParticleContainer{T} <: AbstractParticleContainer
    vals::Vector{T}
```
That could be used with arbitrary particle types.

donhausk:
> I left two comments. Solid work, needs a little bit of tidying up as far as whitespace is concerned.
> I'm also not sure about the direction this package is meant to go -- is this meant to be a package that stands alone but can be incorporated into Turing (like AdvancedHMC) or is it a package that requires Turing to function? Some people may actually really like to be able to just hand some likelihood function to these particle samplers without building a model in Turing.
> If the goal is to have something separate which Turing extends, I highly recommend reading through the interface guide here (sorry it's a little out of date, the Turing.Interface module should now be AbstractMCMC) to see how to implement the various interface methods in AbstractMCMC.

Thanks for your comment. The goal of this package is of course to have something similar to AdvancedHMC for particle filters. However, so far it was the main goal to extract the particle filter from turing and create an own package. There is no proper interface yet. It would definitely make sense to invest some time on creating it! Unfortunately, I do have other projects in the pipeline and therefore, most likely, will not be able to work on this for the moment. 

My aim is to make a functional first pr such that everyone has a solid basis and can contribute to the package. Thanks for your comments! I will of course apply these changes.

src/samplers.jl
donhausk: Could we not make this step in a second PR?

src/samplers.jl
donhausk: Could we not make this step in a second PR?

src/tasks.jl
donhausk: Thanks! I will change it

src/tasks.jl
donhausk: No, they do not use it, however, my goal is to hide the Trace struct form Turing. Creating an empty particle container and filling it using push allows us to change the structure of the trace struct without the need for adapting the Turing code. This might be important when changing the consumer/producer setting to the more recent channels https://github.com/TuringLang/Turing.jl/issues/466.

src/samplers.jl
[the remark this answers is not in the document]
donhausk: I see, thanks! Is this not missleading because there is already a SMC sampler defined in Turing. Additionally, these sampler to not have the same structure as the Turing Sampler.

src/tasks.jl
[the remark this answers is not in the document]
donhausk: Good point!

src/tasks.jl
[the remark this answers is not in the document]
donhausk: True, I have changed it...

src/tasks.jl
devmotion: I think, this question does not depend on the implementation in Turing? The current version of Turing doesn't use any empty initialization anymore. No matter how particles are implemented and which structs are used, I guess usually one create an array of particles first if there are multiple particles (so there's not really any need for the `push!` function that adds `N` particles). If at some points particle have to be added dynamically (which is not the case in the current implementation in Turing), the `push!(::ParticleContainer, ::Particle)` function should be sufficient.

src/samplers.jl
devmotion: Sure

src/samplers.jl
devmotion: Sure

src/tasks.jl
donhausk: I see, thanks for the comment. I will change it.

PR #11, opened 2020-08-18
Particle sampler implementation

denainjs:
Library for particle samplers in Turing.

PR #14, opened 2020-11-30
Move some files

devmotion:
This PR moves the Turing files, apart from the additional algorithms which are outdated, to the src and test folders.

I had another look at https://github.com/TuringLang/AdvancedPS.jl/pull/11 and unfortunately I'm not sure anymore if and how the PR could be updated in the best way. The problem is that the PR is based on an older version of Turing (the particle filter implementation was changed after it was created) and contains many different changes, e.g. files are split and renamed, docstrings are changed, arguments are renamed, and some implementations are modified. Therefore probably it will be a bit tricky to rebase or merge this PR, and to clearly separate the important changes.


yebai:
> I had another look at #11 and unfortunately I'm not sure anymore if and how the PR could be updated in the best way. The problem is that the PR is based on an older version of Turing (the particle filter implementation was changed after it was created) and contains many different changes, e.g. files are split and renamed, docstrings are changed, arguments are renamed, and some implementations are modified. Therefore probably it will be a bit tricky to rebase or merge this PR, and to clearly separate the important changes.

I am more comfortable with a fresh start instead of merging #11. As far as I know, the primary goal of #11 is to remove the dependency on `Turing.jl` by replacing the `produce`/`consume` paradigm with an `iterator` paradigm. Since #11 is not complete, such changes can probably be done in a separate effort rather than rebasing #11, whichever takes less effort.

yebai (approved):

PR #15, opened 2020-12-02
Extract resampling methods

devmotion:
This PR extracts and tests the resampling methods.


devmotion:
In my opinion, in a subsequent PR they should be changed to include explicit RNGs (somehow this is still missing...). I just wanted to only extract things in this PR.

yebai (approved):
It looks all good! Happy to merge as-is given that there are no functional changes.

PR #16, opened 2020-12-02
Rename Turing/src/contrib/inference/AdvancedSMCExtensions.jl to src/c…

yebai:
…ontrib/AdvancedSMCExtensions.jl


devmotion (approved):
The file is not very useful currently though, it seems to be based on an earlier iteration of Turing (maybe even before AbstractMCMC existed?).

yebai:
> The file is not very useful currently though, it seems to be based on an earlier iteration of Turing (maybe even before AbstractMCMC existed?).

yes, we kept algorithms in this file as add-on features while @cpfiffer was transferring Turing into the `AbstractMCMC` interface. Maybe we can keep the file as-is and decide whether to adapt/remove it later?

devmotion:
Sure, I just wanted to point out that it's not really properly integrated like the other files, neither here nor in Turing. But my intention when transferring it from Turing was also that at some point it could be updated.

PR #19, opened 2020-12-05
Add RNG to resampling methods

devmotion:


Project.toml
```diff
@@ -1,11 +1,12 @@
 name = "AdvancedPS"
 uuid = "576499cb-2369-40b2-a588-c64705576edc"
 authors = ["TuringLang"]
-version = "0.1.0"
+version = "0.2.0"
```
devmotion, line 4: Officially, this is not breaking since currently nothing is exported. However, I know that this will break Turing since we call some of the internal methods without RNG, so I thought it would be safer to declare it as breaking.

yebai (approved):
Looks good!

PR #20, opened 2020-12-09
SMC and PG algorithms

devmotion:
This removes Turing-specific parts from the SMC and PG algorithms. The (quite limited) tests copied from Turing pass with a callable model but honestly there might very well be some problems with how these rudimentary models and algorithms handle samples and how they are reset.

Some things I noticed:
- One of the main problems/question is how to return samples/trajectories and if/how to provide a special model structure (so far one has to implement a custom model)
- As mentioned before, currently models (also in Turing) only use the global RNG. This is problematic for parallel sampling or just whenever someone passes a custom RNG to `sample`, hence it might be good to instead implement models as `function model(rng::AbstractRNG) ... end` and to make the RNG explicit in `Trace`. I assume that any in general it would be easier to replay trajectories if particles/traces would save the unmutated RNG with which the model was called.

### DIFF
--- a/test/smc.jl
+++ b/test/smc.jl
@@ -1,179 +1,153 @@
-using Turing, Random, Test
-using Turing.Core: ResampleWithESSThreshold
-using Turing.Inference: getspace, resample_systematic, resample_multinomial
+@testset "smc.jl" begin
+    @testset "SMC constructor" begin
+        sampler = AdvancedPS.SMC(10)
+        @test sampler.nparticles == 10
+        @test sampler.resampler == AdvancedPS.ResampleWithESSThreshold()
 
-using Random
+        sampler = AdvancedPS.SMC(15, 0.6)
+        @test sampler.nparticles == 15
+        @test sampler.resampler ===
+            AdvancedPS.ResampleWithESSThreshold(AdvancedPS.resample_systematic, 0.6)
 
-dir = splitdir(splitdir(pathof(Turing))[1])[1]
-include(dir*"/test/test_utils/AllUtils.jl")
+        sampler = AdvancedPS.SMC(20, AdvancedPS.resample_multinomial, 0.6)
+        @test sampler.nparticles == 20
+        @test sampler.resampler ===
+            AdvancedPS.ResampleWithESSThreshold(AdvancedPS.resample_multinomial, 0.6)
 
-@testset "SMC" begin
-    @turing_testset "constructor" begin
-        s = SMC()
-        @test s.resampler == ResampleWithESSThreshold()
-        @test getspace(s) === ()
-
-        s = SMC(:x)
-        @test s.resampler == ResampleWithESSThreshold()
-        @test getspace(s) === (:x,)
-
-        s = SMC((:x,))
-        @test s.resampler == ResampleWithESSThreshold()
-        @test getspace(s) === (:x,)
-
-        s = SMC(:x, :y)
-        @test s.resampler == ResampleWithESSThreshold()
-        @test getspace(s) === (:x, :y)
-
-        s = SMC((:x, :y))
-        @test s.resampler == ResampleWithESSThreshold()
-        @test getspace(s) === (:x, :y)
-
-        s = SMC(0.6)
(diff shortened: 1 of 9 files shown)


cpfiffer:
>     * One of the main problems/question is how to return samples/trajectories and if/how to provide a special model structure (so far one has to implement a custom model)

I don't have a good enough sense at the moment for some kind of model type. My sense is that it doesn't seem to be needed right now -- adding typing wouldn't help too much. Maybe down the road if `AbstractPPL` becomes a thing we can link into that.



>     * As mentioned before, currently models (also in Turing) only use the global RNG. This is problematic for parallel sampling or just whenever someone passes a custom RNG to `sample`, hence it might be good to instead implement models as `function model(rng::AbstractRNG) ... end` and to make the RNG explicit in `Trace`. I assume that any in general it would be easier to replay trajectories if particles/traces would save the unmutated RNG with which the model was called.

Yeah, that sounds like a bug to me. Including the RNG in the `Trace` sounds good to me, though I wonder how big they are? Is it too onerous to include them in every trace?

cpfiffer (approved):
Looks perfectly fine to me. I had some discussion items above, but I don't think they're core to this PR and can probably be ignored for the moment.

test/smc.jl
```diff
@@ -1,179 +1,153 @@
...
-        @test getspace(s) === (:x,)
+            NormalModel() = new()
+        end
 
-        s = SMC(:x, :y)
-        @test s.resampler == ResampleWithESSThreshold()
-        @test getspace(s) === (:x, :y)
+        function (m::NormalModel)()
+            # First latent variable.
+            m.a = a = rand(Normal(4, 5))
```
yebai, line 33: Maybe consider passing `rng` into the model?

test/smc.jl
```diff
@@ -1,179 +1,153 @@
...
 
-        s = SMC((:x, :y))
-        @test s.resampler == ResampleWithESSThreshold()
-        @test getspace(s) === (:x, :y)
+            # First observation.
+            produce(logpdf(Normal(a, 2), 3))
```
yebai, line 36: Maybe consider defining a helper function `observe(dist, x) = produce(dist, x))` so that we don't need to explicitly export the `produce` API from libtask?

test/smc.jl
```diff
@@ -1,179 +1,153 @@
...
+            m.b = b = rand(Normal(a, 1))
+            if a >= 4
+                produce(logpdf(Normal(b, 2), 1.5))
+            end
+        end
 
-        s = SMC(resample_systematic, (:x,))
-        @test s.resampler === resample_systematic
-        @test getspace(s) === (:x,)
+        @test_throws ErrorException sample(FailSMCModel(), AdvancedPS.SMC(100))
```
yebai, line 62: looks good!

yebai (commented):
thanks @devmotion, this looks good overall. I left a few minor comments below. The tests look reasonable to me, but I agree we can add a few more tests in future PRs.

test/smc.jl
devmotion: I tried this in my local branch when I realized that currently RNGs are not handled corrected by `Trace`. However, I reverted this change when I realized that this would require significant changes to `Trace`.

test/smc.jl
devmotion: Good idea!

devmotion: Done.

test/smc.jl
devmotion: See also https://github.com/TuringLang/AdvancedPS.jl/issues/21#issuecomment-813342838.

devmotion:
Tests fail since Libtask_jll is not compatible with Julia 1.6 yet (more information can be found in the discussion in https://github.com/TuringLang/Turing.jl/issues/1554).

Maybe we should just run tests with Julia 1.5 until it is fixed?

test/smc.jl
yebai: Happy to address this in a separate PR (#23)

yebai (approved):

yebai:
Many thanks @devmotion!

PR #23, opened 2021-04-06
Fix forkr - Handle rng in Trace

FredericWantiez:
Still very early, just opening it to track changes/ideas.


devmotion:
Maybe it would be good to merge https://github.com/TuringLang/AdvancedPS.jl/pull/20 first?

FredericWantiez:
Agree, we need to build on #20 for this fix.

yebai:
> Agree, we need to build on #20 for this fix.

I've merged #20.

src/rng.jl
```diff
@@ -0,0 +1,31 @@
...
+
+# Dispatch to Random, keep track of numbers generated
+function Random.rand!(rng::TracedRNG, values...)
+    res = Random.rand(rng.rng, values...)
+    n = length(res)
+    inc_count!(rng, n)
+    return res
+end
+
+inc_count!(rng::TracedRNG) = inc_count!(rng, 1)
```
devmotion, line 27: To me it still seems that [counter-based PRNGs](https://en.wikipedia.org/wiki/Counter-based_random_number_generator_(CBRNG)) might be better here: they would allow to (re-)generate the n-th random number without having to generate the first n-1 random numbers.

src/smc.jl
```diff
@@ -38,10 +38,10 @@ function AbstractMCMC.sample(
 )
     if !isempty(kwargs)
         @warn "keyword arguments $(keys(kwargs)) are not supported by `SMC`"
-    end
+			end
 
     # Create a set of particles.
-    particles = ParticleContainer([Trace(model) for _ in 1:sampler.nparticles])
+    particles = ParticleContainer([Trace(model, rng) for _ in 1:sampler.nparticles])
```
devmotion, line 44: I think it is crucial to not use the user-provided RNG directly if one wants to keep track of the number of generated random numbers in order to achieve reproducibility: if the user updates the state of the RNG outside of the trace (e.g., by generating other random numbers, in particular if it is just the default RNG) the sequence of random numbers in the trace can not be reproduced anymore.

src/rng.jl
devmotion: I think the implementation is actually in https://github.com/sunoru/Random123.jl.

devmotion: An example on the GPU: https://juliafolds.github.io/FoldsCUDA.jl/dev/examples/monte_carlo_pi/

yebai: I like the idea, however, I would suggest that we keep this as future work. For this PR, we probably want to pass `rng` around all necessary places first, so that we can fix `forkr` next.

src/smc.jl
FredericWantiez: Good point, maybe we can make a copy of the rng and pass it to Trace ?

src/rng.jl
devmotion: :+1: 

I just noticed that DiffEqNoiseProcesses also uses splittable counter-based RNGs: https://github.com/SciML/DiffEqNoiseProcess.jl/blob/master/src/noise_interfaces/virtual_brownian_tree_interface.jl and https://diffeq.sciml.ai/stable/features/noise_process/#VirtualBrownianTree

FredericWantiez:
@yebai I think there's an issue with using fork and passing the RNG to Trace. When we fork the particles during the resampling step, the RNG is set to the RNG we used in the original Trace definition, all children particles will use it and we won't be able to reconstruct the stream.

Also, on the side, when do we populate the reference particle when we resample in PG ?

yebai (changes requested):
We'll need a mechanism to split the random stream during the resampling step. The splitting operation 

1. allows the children particles to share the random history before the splitting point, 
2. make the random streams produce independent random numbers after the splitting point.

> Also, on the side, when do we populate the reference particle when we resample in PG ?

I'm not sure I understand the question. The reference particle is incrementally replayed during the forward propagation of a conditional SMC step. If the reference particle has multiple children at a resampling step, one child will inherit the random stream and keep replaying, the other children will split the random stream and start producing new random numbers. 

Also, happy to discuss more when we meet.

devmotion:
> We'll need a mechanism to split the random stream during the resampling step. The splitting operation
> 
>     1. allows the children particles to share the random history before the splitting point,
> 
>     2. make the random streams produce independent random numbers after the splitting point.

Isn't this exactly what splittable RNGs are designed for (e.g. https://github.com/google/jax/blob/master/design_notes/prng.md#design, https://dl.acm.org/doi/abs/10.1145/2660193.2660195, and https://publications.lib.chalmers.se/records/fulltext/183348/local_183348.pdf)? The last reference also discusses why naive approaches for splitting linear PRNGs fail or make it hard to assess the quality of the returned random numbers (I am not sufficiently familiar with this area to determine if these all are valid concerns).

yebai:
yes, we are planning to use splittable RNGs.

FredericWantiez:
Looked at the random module in tensorflow recently, they also use counter-based rng and the split api: https://github.com/tensorflow/tensorflow/blob/v2.5.0/tensorflow/python/ops/stateful_random_ops.py#L933-L981

We could do something similar:
```julia
mutable struct TracedRNG{R,T<:Random123.AbstractR123{R}}
    count::Int
    rng::T
    keys::Array{R}
end


TracedRNG(r::Random123.AbstractR123) = TracedRNG(0, r, [r.key])


function split(r::TracedRNG, n::Integer)
    rngs = map(1:n) do i
        new_seed = hash(r.rng.key, r.rng.ctr1+i)
        new_rng = TracedRNG(r.count, Philox2x(new_seed), [r.keys..., new_seed])
        return new_rng
    end
end

rng = TracedRNG(Philox2x())
a, b = split(rng, 2)  
```                                                                                                                                                                                                                                                                                                                            
To replay the random numbers we should be able to cycle through the keys and reset the counters

FredericWantiez:
@yebai I simplified the API slightly, we don't need to track both the model counter and the rng counters. We can use the model counter if every particle has an unique key/id. Fixed the naming and the resampling step that was wrong. Also added a few tests. I can add some more, don't think we catch all the edge cases.

yebai (changes requested):
Thanks @FredericWantiez - excellent work. I have gone through the `rng.jl` file and made some comments along the process. I'll take a look at other parts slightly later. 

In particular,  I find the function pair `split` and `update_rng!` is slightly confusing while reading the code. Given that the `split` and `update_rng!` functions are always used in pairs, I wonder whether we can consider merging these functions into `split!(rng::TracedRNG)::TracedRNG` for better clarity. This `split!` function should always split a given RNG once, mutate it in place and return the mutated RNG. If we need to split N-times, we simply call `split` N-times repetitively. Below is an example implementation of the `split!` function: 


```julia
function split!(rng::TracedRNG{T}) where {T}
    key = hash(r.rng.key, convert(UInt, r.rng.ctr1)
    seed!(rng, key)
    return set_counter!(rng.rng, rng.count + 1) # rng.count + 1 is used instead of rng.count
end
```

In order to ensure that calling `split!` repetitively with the same RNG will return new RNGs with different keys, we increase the internal counter by one every time a split is performed.  This way, the next time when `split!` is called on the same RNG, it will get a different key because `r.rng.ctr1` will be different. 

A side question: do we actually need to set `rng.rng.ctr1` to `rng.count` when performing a split? If so, is there a reference on this?

FredericWantiez:
@devmotion @yebai Fixed the last few thing we discussed last week, let me know what you think.

src/rng.jl
```diff
@@ -0,0 +1,104 @@
+using Random123
+using Random
+using Distributions
```
devmotion, line 3: Could these be moved to src/AdvancedPS.jl?

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+
+import Base.rand
+import Random.seed!
+import Random123: set_counter!
```
devmotion, line 7: IMO it is safer and more readable to not import the extended functions but to define the new methods with
```julia
Base.rand(...) = ...
```
etc.

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+
+# Default RNG type for when nothing is specified
+_BASE_RNG = Philox2x
```
devmotion, line 10:
```suggestion
const _BASE_RNG = Philox2x
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+
+"""
+    TracedRNG{R,T}
+
+Wrapped random number generator from Random123 to keep track of random streams during model evaluation
+"""
+mutable struct TracedRNG{T} <:
+               Random.AbstractRNG where {T<:(Random123.AbstractR123{R} where {R})}
```
devmotion, line 18:
```suggestion
mutable struct TracedRNG{T<:Random123.AbstractR123} <: Random.AbstractRNG
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+    "Model step counter"
+    count::Int
+    "Inner RNG"
+    rng::T
+    "Array of keys"
+    keys
```
devmotion, line 24: Make this concretely typed as well?

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+
+"""
+    TracedRNG()
+
+Create a default TracedRNG
+"""
+function TracedRNG()
+    r = _BASE_RNG()
+    return TracedRNG(r)
+end
```
devmotion, line 45: A bit simpler:
```suggestion
"""
    TracedRNG(r::Random123.AbstractR123=AdvancedPS._BASE_RNG())

Create a `TracedRNG` with `r` as the inner RNG. 
"""
function TracedRNG(r::Random123.AbstractR123=_BASE_RNG())
    set_counter!(r, 0)
    return TracedRNG(1, r, typeof(r.key)[])
end
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+
+# Connect to the Random API
+Random.rng_native_52(rng::TracedRNG{U}) where {U} = Random.rng_native_52(rng.rng)
+Base.rand(rng::TracedRNG{U}, ::Type{T}) where {U,T} = Base.rand(rng.rng, T)
```
devmotion, line 49: The type parameters are not needed:
```suggestion
Random.rng_native_52(rng::TracedRNG) = Random.rng_native_52(rng.rng)
Base.rand(rng::TracedRNG, ::Type{T}) where {T} = rand(rng.rng, T)
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+
+"""
+    split(key::Integer, n::Integer)
+
+Split key into n new keys
+"""
+function split(key::Integer, n::Integer=1) where {T}
```
devmotion, line 56:
```suggestion
function split(key::Integer, n::Integer=1)
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+    return map(i -> hash(key, convert(UInt, i)), 1:n)
```
devmotion, line 57: A bit simpler, I think:
```suggestion
    return [hash(key, i) for i in UInt(1):UInt(n)]
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+end
+
+# Connect to the Random API
+Random.rng_native_52(rng::TracedRNG{U}) where {U} = Random.rng_native_52(rng.rng)
+Base.rand(rng::TracedRNG{U}, ::Type{T}) where {U,T} = Base.rand(rng.rng, T)
+
+"""
+    split(key::Integer, n::Integer)
+
+Split key into n new keys
```
devmotion, line 54:
```suggestion
Split `key` into `n` new keys.
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+    r = _BASE_RNG()
+    return TracedRNG(r)
+end
+
+# Connect to the Random API
+Random.rng_native_52(rng::TracedRNG{U}) where {U} = Random.rng_native_52(rng.rng)
+Base.rand(rng::TracedRNG{U}, ::Type{T}) where {U,T} = Base.rand(rng.rng, T)
+
+"""
+    split(key::Integer, n::Integer)
```
devmotion, line 52:
```suggestion
    split(key::Integer, n::Integer=1)
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+function split(key::Integer, n::Integer=1) where {T}
+    return map(i -> hash(key, convert(UInt, i)), 1:n)
+end
+
+"""
+    load_state!(r::TracedRNG, seed)
+
+Load state from current model iteration. Random streams are now replayed
+"""
+function load_state!(rng::TracedRNG{T}) where {T}
```
devmotion, line 65:
```suggestion
function load_state!(rng::TracedRNG)
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+    Random.seed!(rng.rng, key)
+    return set_counter!(rng.rng, 0)
+end
+
+"""
+    update_rng!(rng::TracedRNG)
+
+Set key and counter of inner RNG to key and the running model step
+"""
+function seed!(rng::TracedRNG{T}, key) where {T}
```
devmotion, line 76:
```suggestion
function seed!(rng::TracedRNG, key)
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+    seed!(rng.rng, key)
+    return set_counter!(rng.rng, 0)
+end
+
+""" 
+    save_state!(r::TracedRNG)
+
+Track current key of the inner RNG
+"""
+function save_state!(r::TracedRNG{T}) where {T}
```
devmotion, line 86:
```suggestion
function save_state!(r::TracedRNG)
```

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+    return push!(r.keys, r.rng.key)
+end
+
+Base.copy(r::TracedRNG{T}) where {T} = TracedRNG(r.count, copy(r.rng), copy(r.keys))
```
devmotion, line 90:
```suggestion
Base.copy(r::TracedRNG) = TracedRNG(r.count, copy(r.rng), copy(r.keys))
```
I am also worried that `copy(r.rng)` is not sufficient to ensure that the copy is decoupled from the original RNG. Maybe it would be safer (albeit more expensive) to use `deepcopy`?

src/rng.jl
```diff
@@ -0,0 +1,104 @@
...
+    set_count!(r::TracedRNG, n::Integer)
+
+Set the counter of the TracedRNG, used to keep track of the current model step
+"""
+set_counter!(r::TracedRNG, n::Integer) = r.count = n
+
+"""
+    inc_counter!(r::TracedRNG, n::Integer=1)
+
+Increase the model step counter by n
```
devmotion, line 102:
```suggestion
Increase the model step counter by `n`.
```

test/smc.jl
```diff
@@ -149,6 +150,34 @@
...
         @test mean(x.logevidence for x in chains_pg) ≈ -2 * log(2) atol = 0.01
     end
+
+    @testset "Replay reference" begin
+        mutable struct Model <: AbstractMCMC.AbstractModel
+            a::Float64
+            b::Float64
+
+            Model() = new()
+        end
```
devmotion, line 160: IIRC one has to define structs at the top level.

src/container.jl
```diff
@@ -48,16 +60,18 @@ end
 # Create new task and copy randomness
 function forkr(trace::Trace)
     newf = reset_model(trace.f)
+    set_count!(trace.rng, 1)
+
     ctask = let f = trace.ctask.task.code
         Libtask.CTask() do
-            res = f()
+            res = f()(trace.rng)
```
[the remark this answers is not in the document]
cpfiffer: Isn't `f` a function generator, here?

src/rng.jl
cpfiffer: Yeah, I'd prefer to see all the imports in `src/AdvancedPS.jl`.

cpfiffer (commented):
I think David's already added most of the things I would add -- in general this seems to be fine with a few minor changes needed.

src/rng.jl
FredericWantiez: Maybe something like this: 
```julia
mutable struct TracedRNG{R, T<:Random123.AbstractR123{R}} <: Random.AbstractRNG
    "Model step counter"
    count::Int
    "Inner RNG"
    rng::T
    "Array of keys"
    keys::Array{R}
end
```

devmotion: `Array{R}` is not a concrete type but e.g. `Vector{R}`, `Matrix{R}` or `Array{R,N}` (with additional type parameter `N`) would be.

test/smc.jl
FredericWantiez: Seems like we define the structs locally in most of `test/smc.jl` but I can move it at the top module.
https://github.com/TuringLang/AdvancedPS.jl/blob/6979bbef862589a2dbe4cad062911409c733333e/test/smc.jl#L66-L98

src/rng.jl
FredericWantiez: My bad, we need `Array{R,N}`. I'll change the types accordingly.

src/rng.jl
```diff
@@ -0,0 +1,84 @@
...
+Random.rng_native_52(rng::TracedRNG) = Random.rng_native_52(rng.rng)
+Base.rand(rng::TracedRNG, ::Type{T}) where {T} = Base.rand(rng.rng, T)
+
+"""
+    split(key::Integer, n::Integer=1)
+
+Split `key` into `n` new keys
+"""
+function split(key::Integer, n::Integer=1)
+    return [hash(key, i) for i in UInt(1):UInt(n)]
```
yebai, line 37:
```suggestion
    T = typeof(key) # Make sure the type of `key` is consistent on W32 and W64 systems. 
    return T[hash(key, i) for i in UInt(1):UInt(n)]
```

yebai:
Excellent work - many thanks @FredericWantiez @devmotion and @cpfiffer! 

Perhaps we can wait for a few days to see whether we miss anything, then make a new release.

src/container.jl
```diff
@@ -1,31 +1,43 @@
-struct Trace{F}
+struct Trace{F,U,N,V<:Random123.AbstractR123{U}}
```
devmotion, line 1: Do we dispatch on `U`, `N` or `V` somewhere? Otherwise (and probably even in this case) this could be simplified to
```julia
struct Trace{F,R<:TracedRNG}
    f::F
    ctask::Libtask.CTask
    rng::R
end
```

src/container.jl
```diff
@@ -81,15 +95,21 @@ Data structure for particle filters
 - normalise!(pc::ParticleContainer)
 - consume(pc::ParticleContainer): return incremental likelihood
 """
-mutable struct ParticleContainer{T<:Particle}
+mutable struct ParticleContainer{T<:Particle,U,N,V<:Random123.AbstractR123{U}}
```
devmotion, line 98: Same here, I guess we could use
```julia
mutable struct ParticleContainer{T<:Particle,R<:TracedRNG}
    vals::Vector{T}
    logWs::Vector{Float64}
    rng::R
end
```
?

src/container.jl
```diff
@@ -170,6 +193,22 @@ function effectiveSampleSize(pc::ParticleContainer)
...
+function update_keys!(pc::ParticleContainer, ref::Union{Particle,Nothing}=nothing)
+    # Update keys to new particle ids
+    nparticles = length(pc)
+    n = ref === nothing ? nparticles : nparticles - 1
+    for i in 1:n
+        pi = pc.vals[i]
+        k = split(pi.rng.rng.key)
+        Random.seed!(pi.rng, k[1])
+    end
+end
```
devmotion, line 210: I guess we could add an empty `return` or `nothing` at the end to avoid that the last seed or rng is returned.

test/container.jl
```diff
@@ -11,7 +11,7 @@
         # Create a resumable function that always returns the same log probability.
         function fpc(logp)
             f = let logp = logp
-                () -> begin
+                (rng) -> begin
```
devmotion, line 14: The brackets could be omitted.

PR #28, opened 2021-05-31
Use JuliaFormatter

FredericWantiez:
- Use @devmotion script for CI 
- Run `format(".")`

### DIFF
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -1,0 +1,33 @@
+name: Format
+
+on:
+  push:
+    branches:
+      # This is where pull requests from "bors r+" are built.
+      - staging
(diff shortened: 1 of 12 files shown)


.github/workflows/Format.yml
```diff
@@ -0,0 +1,33 @@
+name: Format
+
+on:
+  push:
+    branches:
+      # This is where pull requests from "bors r+" are built.
+      - staging
+      # This is where pull requests from "bors try" are built.
+      - trying
+      # Build the master branch.
```
devmotion, line 10: We don't use bors in AdvancedPS, so these lines are not needed:
```suggestion
```

devmotion (approved):
Looks good, just had a minor suggestion.

PR #29, opened 2021-06-13
Propagate ref particle in PG

FredericWantiez:
I think we need to propagate the reference particle in PG but ```sweep!``` is missing a ```ref``` parameter

### DIFF
--- a/src/smc.jl
+++ b/src/smc.jl
@@ -114,7 +114,7 @@ function AbstractMCMC.step(
     particles = ParticleContainer(x)
 
     # Perform a particle sweep.
-    logevidence = sweep!(rng, particles, sampler.resampler)
+    logevidence = sweep!(rng, particles, sampler.resampler, particles.vals[nparticles])
 
     # Pick a particle to be retained.
     newtrajectory = rand(rng, particles)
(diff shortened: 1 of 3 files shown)


devmotion:
Isn't this handled in https://github.com/TuringLang/AdvancedPS.jl/blob/176640e114f308874aa92e6ea24c6579642d57ca/src/smc.jl#L104-L113?

FredericWantiez:
Was reading through ```resample_propagate```, looks we need to set ```ref``` to replace the selected particle in the last spot:
https://github.com/TuringLang/AdvancedPS.jl/blob/master/src/container.jl#L225-L229

src/smc.jl
```diff
@@ -114,7 +114,7 @@ function AbstractMCMC.step(
     particles = ParticleContainer(x)
 
     # Perform a particle sweep.
-    logevidence = sweep!(rng, particles, sampler.resampler)
+    logevidence = sweep!(rng, particles, sampler.resampler, particles.vals[nparticles])
```
devmotion, line 117: Should also fix https://github.com/TuringLang/Turing.jl/blob/57e0512e3b63e4622c755dc6d933b51eb76d4483/src/inference/AdvancedSMC.jl#L304 when a new release of AdvancedPS is available.

yebai (approved):

yebai:
Many thanks @FredericWantiez  - looks correct to me!

PR #31, opened 2021-09-25
TracedRNG type

FredericWantiez:
Change type signature and format from review in #23


yebai (approved):
thanks @FredericWantiez, looks good to me!

devmotion (approved):

PR #32, opened 2021-11-03
Particle Gibbs with ancestor resampling

FredericWantiez:
Adding ancestor resampling for state space models, this works I think but it's mostly to discuss the API for the model/sampler. In this I assume something like:

```julia
mutable struct NonLinearTimeSeries <: AdvancedPS.AbstractStateSpaceModel
    X::Vector{Float64}
    θ::θ
    T::Int32
    NonLinearTimeSeries(θ::θ) = new(Vector{Float64}(), θ, 1)
    NonLinearTimeSeries() = new(Vector{Float64}(), θ₀, 1)
end

function AdvancedPS.initialization(m::NonLinearTimeSeries)
    return Normal(0, m.θ.σ^2)
end

function AdvancedPS.transition(m::NonLinearTimeSeries, state)
    return Normal(m.θ.ν + m.θ.α * state, m.θ.σ^2)
end

function AdvancedPS.observation(m::NonLinearTimeSeries, state)
    return m.T > length(y) ? nothing : logpdf(Normal(0, exp(state)), y[m.T])
end
```

**Edit**
The model API changed slightly to make the step and state explicit in the `initialization/observation` functions (needed for the ancestor sampling step)

```julia
mutable struct NonLinearTimeSeries <: AdvancedPS.AbstractStateSpaceModel
    X::Vector{Float64}
    θ::θ
    NonLinearTimeSeries(θ::θ) = new(Vector{Float64}(), θ, 1)
    NonLinearTimeSeries() = new(Vector{Float64}(), θ₀, 1)
end

AdvancedPS.initialization(m::NonLinearTimeSeries) = Normal(0, m.θ.σ^2)
AdvancedPS.transition(m::NonLinearTimeSeries, state, step) = Normal(m.θ.ν + m.θ.α * state, m.θ.σ^2)
AdvancedPS.observation(m::NonLinearTimeSeries, state, step) = logpdf(Normal(0, exp(state)), y[step])
AdvancedPS.isdone(m::NonLinearTimeSeries, step) = step > 3 # Stop the state machine
```


src/model.jl
```diff
@@ -1,3 +1,20 @@
+""" Abstract type for an abstract model formulated in the state space form
+"""
+abstract type AbstractStateSpaceModel <: AbstractMCMC.AbstractModel end
```
devmotion, line 3: Should this be in AdvancedPS?

src/model.jl
```diff
@@ -1,3 +1,20 @@
...
+    ctask::Libtask.CTask
+    rng::R
+end
+
+struct SSMTrace{F<:AdvancedPS.AbstractStateSpaceModel,R<:TracedRNG}
+    f::F
+    rng::R
+end
+
+const Particle = Union{Trace,SSMTrace}
```
devmotion, line 16: This does not seem very extendable. And leads to `Particle` being an abstract type instead of an alias of `Trace`. This seems dangerous both for internal implementations and will likely break downstream code that used the alias.

I think a better alternative might be to change `Trace` to
```julia
struct Trace{F,R<:TracedRNG}
    f::F
    rng::R
end
```
and integrate `ctask` in `f` (probably breaking). Then one could just dispatch on type parameter `F`, and e.g. define
```julia
const SSMTrace{R} = SSMTrace{<:AbstractStateSpaceModel,R}
```
(an abstract type!) as alias for dispatching.

src/pgas.jl
```diff
@@ -0,0 +1,201 @@
...
+Must return `nothing` if it consumed all the data
+"""
+function observation end
+
+"""
+    inc_step!(model::AbstractStateSpaceModel, n::Int=1)
+
+Increase the internal model counter by `n`. Used to keep track of the current timestep when replaying
+"""
+inc_step!(m::AbstractStateSpaceModel, n::Int=1) = m.T += n
```
devmotion, line 28: So every `AbstractStateSpaceModel` is supposed to have a field `T` (BTW maybe use a more descriptive name)? This seems quite specific, maybe better to define this only for the concrete model type that has such a field?

devmotion: This applies to all definitions with `.T` below

src/pgas.jl
```diff
@@ -0,0 +1,201 @@
...
+function transition_logweight(particle::SSMTrace, x)
+    return Distributions.logpdf(transition(particle.f, previous_state(particle.f)), x)
+end
+
+"""
+    get_ancestor_logweights(pc::ParticleContainer{F,R}, x) where {F<:SSMTrace,R}
+
+Get the ancestor log weights for each particle in `pc`
+"""
+function get_ancestor_logweights(pc::ParticleContainer{F,R}, x) where {F<:SSMTrace,R}
```
devmotion, line 67:
```suggestion
function get_ancestor_logweights(pc::ParticleContainer{<:SSMTrace}, x)
```

src/pgas.jl
```diff
@@ -0,0 +1,201 @@
...
+    model = particle.f
+    if !isref
+        if model.T == 1
+            new_state = rand(particle.rng, initialization(model)) # Generate initial state, maybe fallback to 0 is initialization is not defined
+        else
+            current_state = model.X[model.T - 1]
+            new_state = rand(particle.rng, transition(model, current_state))
+        end
+    else
+        new_state = model.X[model.T <= length(model.X) ? model.T : end] # Hacky, reached beyond last time step
```
devmotion, line 96: Throw an error instead?

src/pgas.jl
```diff
@@ -0,0 +1,201 @@
...
+    return score
+end
+
+function truncate!(model::AbstractStateSpaceModel)
+    step = model.T <= length(model.X) ? model.T - 1 : length(model.X)
+    return model.X = model.X[1:step]
+end
+
+function fork(trace::SSMTrace, isref::Bool)
+    model = deepcopy(trace.f)
```
devmotion, line 118: I assume this might become memory intensive?

src/pgas.jl
```diff
@@ -0,0 +1,201 @@
...
+function forkr(trace::SSMTrace)
+    Random123.set_counter!(trace.rng, 1)
+    reset!(trace.f) # Reset model count, we could probably use the TracedRNG inner counter 
+    return SSMTrace(deepcopy(trace.f), trace.rng)
+end
+
+function update_ancestor!(ref::SSMTrace, pc::ParticleContainer)
+    current_step(ref.f) == 1 && return nothing
+
+    particles = collect(pc.vals)
```
devmotion, line 132: Is it necessary to `collect` them?

src/pgas.jl
```diff
@@ -0,0 +1,201 @@
...
+    return ref.f.X[ref.f.T - 1] = previous_state(ancestor.f)
+end
+
+function resample_propagate!(
+    rng::Random.AbstractRNG,
+    pc::ParticleContainer{T,R},
+    randcat=resample_systematic,
+    ref::Union{Particle,Nothing}=nothing;
+    weights=getweights(pc),
+) where {T<:SSMTrace,R}
```
devmotion, line 146:
```suggestion
    pc::ParticleContainer{<:SSMTrace},
    randcat=resample_systematic,
    ref::Union{Particle,Nothing}=nothing;
    weights=getweights(pc),
)
```

src/pgas.jl
```diff
@@ -0,0 +1,201 @@
...
+
+    # check that weights are not NaN
+    @assert !any(isnan, weights)
```
devmotion, line 149: Is this needed? It seems expensive (have to go through all weights in every resampling step) and I assume we would just end up with `NaN` or some other error anyway.

src/model.jl
cpfiffer: I think this should inherit from https://github.com/SciML/DifferenceEquations.jl/blob/14a50c6b6abacbd22e93122f7934b0f787ca721f/src/DifferenceEquations.jl#L11. I'm currently adding in all the nonlinear/arbitrary state space methods so this should be sufficient.

cpfiffer: Though we may need a wrapper that satisfies `<: AbstractMCMC.AbstractModel` anyway to meet the AbstractMCMC requirements

src/pgas.jl
cpfiffer: If we want to plug this into DifferenceEquations.jl, we might use `tspan` instead or otherwise include some kind of subsetting convenience function `subset(prob::StatespaceModel, (lower_t, upper_t))`.

src/model.jl
devmotion: It's a bit unfortunate that SciMLBase isn't as lightweight as one might assume and DifferenceEquations also depends on Optim (which itself depends on ForwardDiff). The large number of dependencies was one of the reasons why we replaced NonlinearSolve (which also depends on SciMLBase and ForwardDiff) with Roots again in Bijectors: https://github.com/TuringLang/Bijectors.jl/pull/202

cpfiffer: The Optim dependency shouldn't be there, I think, that's my bad. I'll remove that.

src/pgas.jl
FredericWantiez: I'll change the name, but we need a way to keep track of the "time" parameter somehow. @cpfiffer not too familiar with DifferenceEquations.jl, what's the convention there ?

src/pgas.jl
FredericWantiez: Yes, this is the main issue with this implementation, we keep copying state vectors around which will probably be quite heavy for long time series

src/pgas.jl
FredericWantiez: Not really, computing the weights would throw an error. That's also in the old resampling step:
https://github.com/TuringLang/AdvancedPS.jl/blob/9f8522f4e2219430255c89f41aa990c2b93104c2/src/container.jl#L230-L231
I can remove these

src/model.jl
FredericWantiez: Agree, missed the issue with downstream using `Particle`. I think unifying `Trace` as you said is probably the best option. It would change a few things in the rest of the package though, not just pgas.

src/model.jl
FredericWantiez: Let me move it, but do we want to add SciMLBase as a dependency ?

src/pgas.jl
FredericWantiez: I get these wrong all the time !

src/pgas.jl
FredericWantiez: No, let me remove it

src/model.jl
yebai: Maybe avoid depending on `SciMLBase` unless there is a very strong reason. In general, we would like to keep the Turing sampler packages lightweight.

cpfiffer: Okay, then let's stick with an internal interface and we'll do the interconnect later.

src/pgas.jl
cpfiffer: Hm, we don't have one yet. Let's stick with what you have here and we'll shoehorn it in later.

yebai:
@FredericWantiez, if you can push your local changes, I’ll take a look later this week.

examples/gaussian-ssm/script.jl
```diff
@@ -1,13 +1,72 @@
...
 using Distributions
-using Libtask: produce
-using StatsFuns
-
 using Plots
 
-x = [1, 2, 3]
-y = [2, 5, 1]
+# Parameters
+θ = @NamedTuple begin
```
yebai, line 8: Maybe consider a more explicit variable name:

```suggestion
Parameters = @NamedTuple begin
```

examples/gaussian-ssm/script.jl
```diff
@@ -1,13 +1,72 @@
...
+q = 0.32  # State variance
+r = 1     # Observation variance
+Tₘ = 300  # Number of observation
+Nₚ = 50   # Number of particles
+Nₛ = 1000 # Number of samples
+seed = 9  # Reproduce everything
+
+θ₀ = θ((a, q, r))
+
+f(x, t) = Normal(a * x, q) # Transition density
```
yebai, line 24: Perhaps take parameters from `m::NonLinearTimeSeries` instead of a global variable? 

```suggestion
f(m::NonLinearTimeSeries, t) = Normal(m.θ.a * m.X[t], model.θ.q) # Transition density
```

examples/gaussian-ssm/script.jl
```diff
@@ -1,13 +1,72 @@
...
+g(x, t) = Normal(x, r)     # Observation density
```
yebai, line 25: Similarily, 
```suggestion
g(m::NonLinearTimeSeries, t) = Normal(m.X[t], m.θ.r)     # Observation density
```

examples/gaussian-ssm/script.jl
```diff
@@ -1,13 +1,72 @@
...
+f₀ = Normal(0, q)          # Initial state density
```
yebai, line 26:
```suggestion
f₀(m:: NonLinearTimeSeries) = Normal(0, m.θ.q)          # Initial state density
```

examples/gaussian-ssm/script.jl
```diff
@@ -1,13 +1,72 @@
...
+
+mutable struct NonLinearTimeSeries <: AdvancedPS.AbstractStateSpaceModel
+    X::Vector{Float64}
+    θ::θ
```
yebai, line 30:
```suggestion
    θ::Parameters
```

examples/gaussian-ssm/script.jl
```diff
@@ -1,13 +1,72 @@
...
+    if t < Tₘ
+        x[t + 1] = rand(rng, f(x[t], t))
+    end
+    y[t] = rand(rng, g(x[t], t))
+end
+
+plot(x; label="x", color=:black)
+plot(y; label="y", color=:black)
+
+# Turing
```
yebai, line 57:
```suggestion
# Setup up a particle Gibbs sampler
```

src/container.jl
```diff
@@ -269,10 +265,10 @@ function resample_propagate!(
             pi = particles[i]
             isref = pi === ref
             p = isref ? fork(pi, isref) : pi
-            nseeds = isref ? ni - 1 : ni
+            rng = isref ? particles[j + 1].rng.rng : p.rng.rng
```
yebai, line 268: There might be a subtle issue here related to `j+1`. If the particle `j+1` has non-zero children, then its `rng` would be used by itself, and maybe a child from the `ref` particle.  That would lead to two particles (one ref's child, one particle `j+1`'s child) sharing the same `rng`. Is that an intended behaviour?

src/model.jl
yebai: @FredericWantiez I agree with @devmotion's suggestion here. It is better to unify these `Trace` types.

src/pgas.jl
```diff
@@ -0,0 +1,136 @@
...
+function previous_state(trace::SSMTrace)
+    return trace.f.X[step(trace) - 1]
+end
+
+"""
+    step(model::AbstractStateSpaceModel)
+
+Return current model step
+"""
+step(trace::SSMTrace) = trace.rng.count
```
yebai, line 45: Would the `ref` particle store all previous `rng`'s? If so, does this `step` function work with the `ref` particle?

src/pgas.jl
```diff
@@ -0,0 +1,136 @@
...
+    !isref && push!(model.X, new_state)
+    inc_counter!(particle.rng) # Increase rng counter, we use it as the model `time` index instead of handling two distinct counters
+
+    return score
+end
+
+function truncate!(particle::SSMTrace)
+    model = particle.f
+    current_step = step(particle)
+    model.X = model.X[1:(current_step - 1)]
```
yebai, line 108: Maybe we can simply mark `model.X[current_step:end]` as un-used so we can avoid unnecessary GC?

src/pgas.jl
```diff
@@ -0,0 +1,136 @@
...
+    particle.rng.keys = particle.rng.keys[1:(current_step - 1)]
+    return model
+end
+
+function fork(particle::SSMTrace, isref::Bool)
```
yebai, line 113: Maybe allow passing in an `old_particle`, so that we can mutate `old_particle` with the values from `particle`. This would allow re-using `old_particle`'s memory. 

```suggestion
function fork(particle::SSMTrace, isref::Bool, old_particle=nothing)
```

src/smc.jl
```diff
@@ -128,3 +137,48 @@ function AbstractMCMC.step(
...
+    # Pick a particle to be retained.
+    trajectory = rand(particles.rng, particles)
+    return PGSample(trajectory, logevidence), PGState(trajectory)
+end
+
+function AbstractMCMC.step(
+    rng::Random.AbstractRNG,
+    model::AbstractStateSpaceModel,
+    sampler::PGAS,
+    state::PGState;
```
yebai, line 163: Maybe consider using the argument signature below, so that we can merge these two `AbstractMCMC.step` implementations? 

```suggestion
    state::Union{PGState,nothing} = nothing;
```

yebai (changes requested):
Many thanks, @FredericWantiez - very good work. I left some minor comments below.

src/pgas.jl
FredericWantiez: Should I address this and the comment on GC in another PR for optimization ?

src/smc.jl
FredericWantiez: I also changed the signature of `PG`, most of the code was duplicated

src/container.jl
FredericWantiez: That was indeed an issue, instead we can pick a RNG a priori and use it to generate keys for the ref particle children. (just to seed the rng, these childrent would still inherit from the reference history)

src/model.jl
FredericWantiez: I moved verything in `model.jl` and unified `Trace`:
https://github.com/TuringLang/AdvancedPS.jl/blob/1f152785a42739afd98d84ebd889b3bde88ef24a/src/model.jl#L1-L34

src/pgas.jl
yebai: Sure, a separate PR sounds good.

src/model.jl
```diff
@@ -1,3 +1,85 @@
+"""
+    TapedModel{F}
+
+State wrapper to hold `Libtask.CTask` model initiated from `f`
+"""
+struct TapedModel{F}
```
devmotion, line 6: I guess this should this also be a subtype of `AbstractModel`?
```suggestion
struct TapedModel{F} <: AbstractMCMC.AbstractModel
```

src/model.jl
```diff
@@ -1,3 +1,85 @@
...
+    f::F
+    ctask::Libtask.TapedTask
```
devmotion, line 8: IIRC `TapedTask` is not a concrete type anymore. Should we use a type parameter here?

src/pgas.jl
```diff
@@ -0,0 +1,142 @@
...
+function previous_state(trace::SSMTrace)
+    return trace.model.X[step(trace) - 1]
+end
+
+function past_idx(trace::SSMTrace)
+    return 1:(step(trace) - 1)
+end
+
+"""
+    step(model::AbstractStateSpaceModel)
```
devmotion, line 45: I guess `step` could be confusing since then one has two different functions `AbstractMCMC.step` and `AdvancedPS.step`.

src/model.jl
```diff
@@ -1,3 +1,85 @@
+"""
+    TapedModel{F}
+
+State wrapper to hold `Libtask.CTask` model initiated from `f`
+"""
+struct TapedModel{F} <: AbstractMCMC.AbstractModel
```
yebai, line 6:
```suggestion
struct GenericModel{F} <: AbstractMCMC.AbstractModel
```

src/model.jl
```diff
@@ -1,3 +1,85 @@
...
+        newtrace = new{F,R}(model, rng)
+        addreference!(model.ctask.task, newtrace)
+        return newtrace
+    end
+end
+
+const Particle = Trace
+
+const SSMTrace{R} = Trace{<:AbstractStateSpaceModel,R}
+const TapedTrace{R} = Trace{<:TapedModel,R}
```
yebai, line 34:
```suggestion
const GenericTrace{R} = Trace{<:TapedModel,R}
```

src/model.jl
```diff
@@ -1,3 +1,90 @@
...
+    # add backward reference
+    newtrace = Trace(new_tapedmodel, trace.rng)
+    addreference!(ctask.task, newtrace)
+    gen_refseed!(newtrace)
+    return newtrace
+end
+
+function gen_refseed!(part::Particle)
+    seed = split(state(part.rng.rng), 1)
+    return set_refseed!(part.rng, seed[1])
```
yebai, line 79: Would this line override the last `key` in `part.rng::TracedRNG`? If so, will it cause issues during replaying particle `part`? 
```suggestion
    return set_refseed!(part.rng, seed[1])
```

src/container.jl
```diff
@@ -263,16 +198,18 @@ function resample_propagate!(
...
         if ni > 0
             # fork first child
             pi = particles[i]
             isref = pi === ref
             p = isref ? fork(pi, isref) : pi
-            nseeds = isref ? ni - 1 : ni
 
-            seeds = split(p.rng.rng.key, nseeds)
-            !isref && Random.seed!(p.rng, seeds[1])
+            key = isref ? refseed(ref.rng) : state(p.rng.rng) # Pick up the alternative rng stream if using the reference particle
```
yebai, line 207:
```suggestion
            key = isref ? safe_get_refseed(ref.rng) : state(p.rng.rng) # Pick up the alternative rng stream if using the reference particle
```

src/container.jl
```diff
@@ -263,16 +198,18 @@ function resample_propagate!(
...
+            nsplits = isref ? ni + 1 : ni # We need one more seed to refresh the alternative rng stream
+            seeds = split(key, nsplits)
+            isref && set_refseed!(ref.rng, seeds[end]) # Refresh the alternative rng stream
```
yebai, line 210:
```suggestion
            isref && safe_set_refseed!(ref.rng, seeds[end]) # Refresh the alternative rng stream
```

src/rng.jl
```diff
@@ -88,6 +88,18 @@ state(rng::Random123.Philox4x) = (rng.key1, rng.key2)
...
+# Add an extra seed to the reference particle keys array to use as an alternative stream 
+# (we don't need to tack this one)
+#
+# We have to be careful when spliting the reference particle. 
+# Since we don't know the seed tree from the previous SMC run we cannot reuse any of the intermediate seed
+# in the TracedRNG container. We might collide with a previous seed and the children particle would collapse
+# to the reference particle. A solution to solve this is to have an extra stream attached to the reference particle 
+# that we only use to seed the children of the reference particle. 
+#
+set_refseed!(rng::TracedRNG{R}, seed::R) where {R} = rng.keys[end] = seed
```
yebai, line 100:
```suggestion
safe_set_refseed!(rng::TracedRNG{R}, seed::R) where {R} = rng.keys[end] = seed
```

src/rng.jl
```diff
@@ -88,6 +88,18 @@ state(rng::Random123.Philox4x) = (rng.key1, rng.key2)
...
+refseed(rng::TracedRNG) = rng.keys[end]
```
yebai, line 101:
```suggestion
safe_get_refseed(rng::TracedRNG) = rng.keys[end]
```

examples/gaussian-ssm/script.jl
```diff
@@ -1,13 +1,75 @@
...
+end
+
+f(m::NonLinearTimeSeries, state, t) = Normal(m.θ.a * state, m.θ.q) # Transition density
+g(m::NonLinearTimeSeries, state, t) = Normal(state, m.θ.r) # Observation density
+f₀(m::NonLinearTimeSeries) = Normal(0, m.θ.q / (1 - m.θ.a^2)) # Initial state density
+
+AdvancedPS.initialization(model::NonLinearTimeSeries) = f₀(model)
+AdvancedPS.transition(model::NonLinearTimeSeries, state, step) = f(model, state, step)
+function AdvancedPS.observation(model::NonLinearTimeSeries, state, step)
+    return logpdf(g(model, state, step), y[step])
```
yebai, line 38: Is there a typo here? It does not seem `y` is in the arguments. 
```suggestion
    return logpdf(g(model, state, step), y[step])
```

src/pgas.jl
```diff
@@ -0,0 +1,151 @@
...
+""" 
+    transition(model::AbstractStateSpaceModel, state, step)
+
+Define the transition density of the State Space Model
+Must return `nothing` if it consumed all the data
+"""
+function transition end
+
+"""
+    observation(model::AbstractStateSpaceModel, state, step)
```
yebai, line 17: Are the variables `state` and `step` redundant with each other, i.e. `state = model.X[step]`? If so, maybe we can remove the extra variable `state` here? Otherwise, perhaps add a comment explaining why. 

EDIT: it seems this is requires in [line](https://github.com/TuringLang/AdvancedPS.jl/pull/32/files#diff-4e0ce155dd587f5f2009d7664a4e1db765a58990171122250646bf9a827aa3c3R102). Let's add a comment explain why it is needed.

yebai:
Many thanks, @FredericWantiez - excellent work! Only a few minor comment/clarification questions remain. Once fixed, we should be ready to go.

src/model.jl
FredericWantiez: I'm hacking around a bit here, but in the current implementation we store T+1 seeds (for a model with T steps) because we iterate T+1 times the model (to get the final un-weighted sample). This last seed is not used in subsequent runs. 

We could also just append a seed to the reference particle here or track it as an optional field in `TracedRNG` here:
https://github.com/TuringLang/AdvancedPS.jl/blob/02a5258c12b2db7f0dc6635311349465ce6d23eb/src/model.jl#L63-L75

yebai: I like tracking it as an optional field in `TracedRNG`. We can simply leave it as uninitialised for non-reference particles. Such a design can improve the clarity, which we should aim for.

FredericWantiez: Agree, it was really hard to follow the logic. Even if I wrote it ...

yebai (approved):

PR #33, opened 2021-11-13
Documentation

FredericWantiez:
Adding doc and examples workflow to the repo. Also a basic writeup for the API. 
The examples are still to-do


FredericWantiez:
Not quite sure why it does not publish across forks but here's a preview 
https://fredericwantiez.github.io/AdvancedPS.jl/previews/PR7/

devmotion:
> Not quite sure why it does not publish across forks but here's a preview https://fredericwantiez.github.io/AdvancedPS.jl/previews/PR7/

This is a limitation of Documenter + Github (also due to a quite restrictive security model of Github actions).

yebai:
Maybe we can merge this PR so that doc previews can build? More improvements can be added in separate PRs. It would make previewing changes possible for this repo.

yebai (approved):

devmotion:
Doc previews will never work for PRs from forks, it's not an issue with setting it up first (but maybe this was clear already).

yebai:
Right - I wasn't aware of that. Once this PR is merged, future PRs can be created from this repo directly. @FredericWantiez I think you can work from this repo directly,  let me know if there is an issue.

docs/make.jl
```diff
@@ -0,0 +1,80 @@
...
+        "api.md",
+        "Examples" => [
+            "example.md",
+            map(
+                (x) -> joinpath("examples", x),
+                filter!(filename -> endswith(filename, ".md"), readdir(EXAMPLES_OUT)),
+            )...,
+        ],
+    ],
+    checkdocs=:exports,
```
devmotion, line 69: Maybe we could use
```suggestion
    strict=true,
    checkdocs=:exports,
```
to ensure that an error is thrown and the action fails if something fails or e.g. is not documented. Otherwise one always has to check the logs.

.github/workflows/TagBot.yml
```diff
@@ -12,4 +12,4 @@ jobs:
       - uses: JuliaRegistries/TagBot@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-
+          ssh: ${{ secrets.DOCUMENTER_KEY }}
```
devmotion, line 15: Was the key and corresponding secret added to the repo?

README.md
```diff
@@ -5,6 +5,20 @@
 [![Coverage](https://coveralls.io/repos/github/TuringLang/AdvancedPS.jl/badge.svg?branch=master)](https://coveralls.io/github/TuringLang/AdvancedPS.jl?branch=master)
 [![Code Style: Blue](https://img.shields.io/badge/code%20style-blue-4495d1.svg)](https://github.com/invenia/BlueStyle)
 
+AdvancedPS provides and efficient implementation of common particle based Monte Carlo samplers using the [AbstractMCMC](https://github.com/TuringLang/AbstractMCMC.jl) interface.
+AdvancedPS is part of the [Turing](https://turing.ml/stable/) ecosystem.
+
+### Installation
+
+Inside the julia REPL
```
devmotion, line 13: IIRC the convention is to use uppercase for the language and lowercase only for the binary:
```suggestion
Inside the Julia REPL
```

README.md
```diff
@@ -5,6 +5,20 @@
 [![Coverage](https://coveralls.io/repos/github/TuringLang/AdvancedPS.jl/badge.svg?branch=master)](https://coveralls.io/github/TuringLang/AdvancedPS.jl?branch=master)
 [![Code Style: Blue](https://img.shields.io/badge/code%20style-blue-4495d1.svg)](https://github.com/invenia/BlueStyle)
 
+AdvancedPS provides and efficient implementation of common particle based Monte Carlo samplers using the [AbstractMCMC](https://github.com/TuringLang/AbstractMCMC.jl) interface.
```
devmotion, line 8:
```suggestion
AdvancedPS provides an efficient implementation of common particle based Monte Carlo samplers using the [AbstractMCMC](https://github.com/TuringLang/AbstractMCMC.jl) interface.
```
Maybe also mention Libtask (until it is replaced)?

docs/make.jl
```diff
@@ -0,0 +1,80 @@
...
+isempty(processes) || success(processes) || error("some examples were not run successfully")
+
+# Building Documenter
+using Documenter
+
+# Print `@debug` statements (https://github.com/JuliaDocs/Documenter.jl/issues/955)
+if haskey(ENV, "GITHUB_ACTIONS")
+    ENV["JULIA_DEBUG"] = "Documenter"
+end
+
```
devmotion, line 49: It's set in the action already:

```suggestion
```

README.md
FredericWantiez: Good point, seems like Libtask is always the pain point when people look at AdvancedPS

.github/workflows/TagBot.yml
FredericWantiez: Yes, generated a new pair and added it. I don't think we have keys at the organization level ?

docs/make.jl
FredericWantiez: Ah, missed that. Yes let's fail instead

FredericWantiez:
@yebai agree, let's get the scaffolding in and I'll move away from my forks

PR #35, opened 2021-12-07
Use tape based Libtask

KDr2:
Run test cases against the tape-based libtask.

We don't support branches in taped libtask now, so I commented out one test case which contains a `if` expression.

All other test cases passed.

### DIFF
--- a/src/container.jl
+++ b/src/container.jl
@@ -56,7 +56,7 @@ function forkr(trace::Trace)
     newf = reset_model(trace.f)
     Random123.set_counter!(trace.rng, 1)
 
-    func = trace.ctask.tf.func
+    func = Libtask.func(trace.ctask)
     ctask = Libtask.CTask(func, trace.rng)
 
     # add backward reference
--- a/src/container.jl
+++ b/src/container.jl
@@ -7,13 +7,7 @@ end
 const Particle = Trace
 
 function Trace(f, rng::TracedRNG)
-    ctask = let f = f
-        Libtask.CTask() do
-            res = f(rng)
-            Libtask.produce(nothing)
-            return res
-        end
-    end
+    ctask = Libtask.CTask(f, rng)
 
     # add backward reference
     newtrace = Trace(f, ctask, rng)
@@ -62,13 +56,8 @@ function forkr(trace::Trace)
     newf = reset_model(trace.f)
     Random123.set_counter!(trace.rng, 1)
 
-    ctask = let f = trace.ctask.task.code
-        Libtask.CTask() do
-            res = f()(trace.rng)
-            Libtask.produce(nothing)
-            return res
-        end
-    end
+    func = trace.ctask.tf.func
+    ctask = Libtask.CTask(func, trace.rng)
 
     # add backward reference
     newtrace = Trace(newf, ctask, trace.rng)
--- a/src/container.jl
+++ b/src/container.jl
@@ -56,7 +56,7 @@ function forkr(trace::Trace)
     newf = reset_model(trace.f)
     Random123.set_counter!(trace.rng, 1)
 
-    ctask = Libtask.CTask(trace.ctask, trace.rng)
+    ctask = Libtask.CTask(newf, trace.rng)
 
     # add backward reference
     newtrace = Trace(newf, ctask, trace.rng)
--- a/src/container.jl
+++ b/src/container.jl
@@ -56,8 +56,7 @@ function forkr(trace::Trace)
     newf = reset_model(trace.f)
     Random123.set_counter!(trace.rng, 1)
 
-    func = Libtask.func(trace.ctask)
-    ctask = Libtask.CTask(func, trace.rng)
+    ctask = Libtask.CTask(trace.ctask, trace.rng)
 
     # add backward reference
     newtrace = Trace(newf, ctask, trace.rng)
--- a/src/model.jl
+++ b/src/model.jl
@@ -8,12 +8,3 @@ function observe(dist::Distributions.Distribution, x)
 end
 
 Libtask.trace_into(::typeof(observe)) = true
-
-#=
-function (instr::Libtask.Instruction{typeof(observe)})()
-    dist = Libtask.val(instr.input[1])
-    x = Libtask.val(instr.input[2])
-    result = Distributions.loglikelihood(dist, x)
-    Libtask.internal_produce(instr, result)
-end
-=#
--- a/src/model.jl
+++ b/src/model.jl
@@ -7,9 +7,13 @@ function observe(dist::Distributions.Distribution, x)
     return Libtask.produce(Distributions.loglikelihood(dist, x))
 end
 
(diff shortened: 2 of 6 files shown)


devmotion:
I think it would be good to make a release of AdvancedPS 0.3 first, update Turing, and add integration tests here. There are many unreleased changes on the master branch of AdvancedPS and it is not clear how well they work with Turing and what changes are needed there. I think it would be better to update the ecosystem in two steps, to avoid mixing up problems or introducing any regressions in the Libtask rewrite (still think it would be better to use a new package :smile:).

src/model.jl
```diff
@@ -6,3 +6,14 @@ Observe sample `x` from distribution `dist` and yield its log-likelihood value.
 function observe(dist::Distributions.Distribution, x)
     return Libtask.produce(Distributions.loglikelihood(dist, x))
 end
+
+function (instr::Libtask.Instruction{typeof(observe)})()
```
devmotion, line 10: Why is this needed? And could we dispatch on the arguments of `observe` if at some point we decide to implement add more `observe` implementations?

KDr2: This is needed because the current tape impl only tracks the direct call to `produce`, the task function calls this function, and this calls `produce`, so `produce` is not directly called by the task function and won't be recorded. This function definition makes `observe` have the same effect as `produce`.

The Instruction type is not parameterized on its input types now, I think it is possible to add this in. I will look into that later.

KDr2:
> still think it would be better to use a new package 😄

I have the same thought, but I am focusing on making sure this approach works now and haven't got any concrete update plans :)

yebai:
> I think it would be good to release AdvancedPS 0.3 first, update Turing, and add integration tests here. There are many unreleased changes on the master branch of AdvancedPS and it is not clear how well they work with Turing and what changes are needed there. 

How about we first make a new release based on the most recent release, but with changes in this PR? This would allow us to support Julia 1.7 first. I think new changes on the master branch and in #32 will likely take more time to push through. 

I have created a new [releases branch](https://github.com/TuringLang/AdvancedPS.jl/tree/releases) from the most recent release. If we merge this PR into the releases brach, we can make a new release there without worrying about the changes on the master branch.

src/container.jl
```diff
@@ -62,13 +56,8 @@ function forkr(trace::Trace)
...
 
-    ctask = let f = trace.ctask.task.code
-        Libtask.CTask() do
-            res = f()(trace.rng)
-            Libtask.produce(nothing)
-            return res
-        end
-    end
+    func = Libtask.func(trace.ctask)
+    ctask = Libtask.CTask(func, trace.rng)
```
yebai, line 60: Maybe create a convenient constructor `Libtask.CTask(t::Ctask, args...)` so we don't need to handle `func` explicitly.

KDr2: Done.

KDr2:
Close in favor of https://github.com/TuringLang/AdvancedPS.jl/pull/36.

yebai:
#36 is a backport to the most recent release, we still need this PR to update the master.

src/model.jl
```diff
@@ -6,3 +6,5 @@ Observe sample `x` from distribution `dist` and yield its log-likelihood value.
 function observe(dist::Distributions.Distribution, x)
     return Libtask.produce(Distributions.loglikelihood(dist, x))
 end
+
+Libtask.trace_into(::typeof(observe)) = true
```
yebai, line 10:
```suggestion
```

test/container.jl
```diff
@@ -12,7 +12,7 @@
         function fpc(logp)
             f = let logp = logp
                 rng -> begin
-                    while true
+                    for _ in 1:100
```
yebai, line 15:
```suggestion
                    while true
```

test/container.jl
```diff
@@ -100,7 +100,7 @@
         function f2(rng)
             t = TArray(Int, 1)
             t[1] = 0
-            while true
+            for _ in 1:100
```
yebai, line 103:
```suggestion
            while true
```

test/smc.jl
```diff
@@ -55,12 +55,12 @@
         function (m::FailSMCModel)(rng::Random.AbstractRNG)
             m.a = a = rand(rng, Normal(4, 5))
             m.b = b = rand(rng, Normal(a, 1))
-            if a >= 4
-                AdvancedPS.observe(Normal(b, 2), 1.5)
-            end
+            # if a >= 4
```
yebai, line 58:
```suggestion
            if a >= 4
```

test/smc.jl
```diff
@@ -55,12 +55,12 @@
...
+            AdvancedPS.observe(Normal(b, 2), 1.5)
+            # end
```
yebai, line 60:
```suggestion
            end
```

test/smc.jl
```diff
@@ -55,12 +55,12 @@
...
         end
 
-        @test_throws ErrorException sample(FailSMCModel(), AdvancedPS.SMC(100))
+        # @test_throws ErrorException sample(FailSMCModel(), AdvancedPS.SMC(100))
```
yebai, line 63:
```suggestion
        @test_throws ErrorException sample(FailSMCModel(), AdvancedPS.SMC(100))
```

yebai (changes requested):

src/container.jl
```diff
@@ -7,13 +7,7 @@ end
...
 
 function Trace(f, rng::TracedRNG)
-    ctask = let f = f
-        Libtask.CTask() do
-            res = f(rng)
-            Libtask.produce(nothing)
-            return res
-        end
-    end
+    ctask = Libtask.CTask(f, rng)
```
FredericWantiez, line 10: Why not match the new Turing interface ? Or that's for a later PR ?
```julia
  if hasfield(typeof(f), :evaluator) # Test whether f is a Turing.TracedModel
      # println(f.evaluator)
      ctask = Libtask.CTask(f.evaluator[1], f.evaluator[2:end]...)
  else # f is a Function, or AdavncedPS.Model
      ctask = Libtask.CTask(f)
  end
```

yebai: yes, we should match that - please feel free to update this PR!

devmotion: Couldn't/Shouldn't this be handled in Turing by implementing `Libtask.CTask(f::TracedModel, rng)`?

src/container.jl
```diff
@@ -62,13 +56,7 @@ function forkr(trace::Trace)
...
     Random123.set_counter!(trace.rng, 1)
 
-    ctask = let f = trace.ctask.task.code
-        Libtask.CTask() do
-            res = f()(trace.rng)
-            Libtask.produce(nothing)
-            return res
-        end
-    end
+    ctask = Libtask.CTask(trace.ctask, trace.rng)
```
yebai, line 59:
```suggestion
    ctask = Libtask.CTask(newf, trace.rng)
```
```suggestion
    ctask = Libtask.CTask(trace.ctask, trace.rng)
```

PR #41, opened 2022-01-15
Seed particles from top-level RNG

FredericWantiez:
With the new model definition, we should be able to seed the particle inner RNG and reproduce the random streams from the RNG provided in `sample`. Currently something like this would produce two different experiments:

```julia
seed = 100

rng = Random.MersenneTwister(seed)
pg = AdvancedPS.PG(3)
model = Model()

chains = sample(rng, model, pg, 2);

Random.seed!(rng, seed)
# Produces different values
chains = sample(rng, model, pg, 2);
```

### DIFF
--- a/src/container.jl
+++ b/src/container.jl
@@ -106,8 +106,15 @@ function ParticleContainer(particles::Vector{<:Particle})
     return ParticleContainer(particles, zeros(length(particles)), TracedRNG())
 end
 
-function ParticleContainer(particles::Vector{<:Particle}, r::TracedRNG)
-    return ParticleContainer(particles, zeros(length(particles)), r)
+function ParticleContainer(particles::Vector{<:Particle}, rng::TracedRNG)
+    return ParticleContainer(particles, zeros(length(particles)), rng)
+end
+
+function ParticleContainer(
+    particles::Vector{<:Particle}, trng::TracedRNG, rng::Random.AbstractRNG
+)
+    pc = ParticleContainer(particles, trng)
+    return seed_from_rng!(pc, rng)
 end
 
 Base.collect(pc::ParticleContainer) = pc.vals
@@ -202,7 +209,7 @@ function update_keys!(pc::ParticleContainer, ref::Union{Particle,Nothing}=nothin
     n = ref === nothing ? nparticles : nparticles - 1
     for i in 1:n
         pi = pc.vals[i]
-        k = split(pi.rng.rng.key)
+        k = split(state(pi.rng.rng))
         Random.seed!(pi.rng, k[1])
     end
     return nothing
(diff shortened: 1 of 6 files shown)


devmotion:
> Currently something like this would produce two different experiments:

Yes, this was observed e.g. in https://github.com/TuringLang/Turing.jl/issues/1731.

But isn't the main issue currently that we never forward the RNG to the Trace and TracedModel constructors, e.g., in https://github.com/TuringLang/Turing.jl/blob/9087412bb574bc83eacd9301f7fa5892a839c666/src/inference/AdvancedSMC.jl#L360-L370 and https://github.com/TuringLang/Turing.jl/pull/1757/files#diff-f2a243e03d83fd80c6af74a78557a37eb7bcf265c509d29b9faa7ec19873b0c5R14?

src/rng.jl
```diff
@@ -59,6 +59,15 @@ function Random.seed!(rng::TracedRNG, key)
     return Random123.set_counter!(rng.rng, 0)
 end
 
+"""
+    gen_seeds(trng::TracedRNG, other::Random.AbstractRNG)
+
+Seed a TracedRNG from another 
+"""
+function gen_seeds(::Type{T}, other::Random.AbstractRNG, n::Integer=1) where {T}
+    return Tuple(rand(other, T, n))
```
devmotion, line 68: This will allocate and be type unstable. I guess it might be easier to just create a sampler in the function above and then loop through the different rngs, without sampling everything upfront. The sampler is optimized for computing a stream, and not only a single, random number: https://docs.julialang.org/en/v1/stdlib/Random/#Random.Sampler

src/smc.jl
```diff
@@ -41,6 +41,7 @@ function AbstractMCMC.sample(
     particles = ParticleContainer(
         [Trace(model, TracedRNG()) for _ in 1:(sampler.nparticles)], TracedRNG()
     )
+    seed_from_rng!(particles, rng)
```
devmotion, line 44: Isn't it a bit easier to allow e.g. to pass `rng` to the `ParticleContainer` constructor?

FredericWantiez:
There are two issues I think, one on the Turing side where we don't forward the RNG but here as well, the previous model was calling on the GLOBAL_RNG implicitly:
```julia
function (m::Model)()
    # First hidden variables.
    m.a = rand(Normal()) # Call the global rng, prior to 1.7 it was not task local
    produce(logpdf(Normal(), 1)) 
end
```
Even if we reference the RNG properly, we still need to set the seeds in the new `TracedRNG`.

src/smc.jl
FredericWantiez: Yes, good point I'll add a constructor

devmotion:
>  the previous model was calling on the GLOBAL_RNG implicitly:

In the tests we already used the definitions `(model::Model)(rng::AbstractRNG) = ...`, so there shouldn't be any implicit `GLOBAL_RNG`?

src/rng.jl
FredericWantiez: Not quite sure I understand why we need to define the sampler, can we just do define the `SamplerType` ?
```julia
for T = (:UInt32, :UInt64)
    for N in (:1, :2, :4)
        @eval begin
            Base.rand(rng::Random.AbstractRNG, ::Random.SamplerType{NTuple{$N, $T}}) = Tuple(rand(rng, $T, $N))
        end
    end
end 
```

FredericWantiez:
Yes that's correct, my bad, got confused with the discussion in TuringLang/Turing.jl#1731

src/rng.jl
devmotion: This would be a quite severe type piracy...

We don't have to construct a sampler but usually it will be more efficient than calling `rand(rng, UInt32)` etc. in a loop. My point was that I don't think you have to sample a tuple at all but can generate random seeds (at least) as efficiently by using something like
```julia
sampler = Random.Sampler(rng, UInt32)
for subrng in subrngs
    Random.seed!(subrng, rand(rng, sampler))
end
```

FredericWantiez: That makes sense, the only issue is some of these `Sampler` are not defined: the seed for a `Random123.Philox4x` is a `Tuple{UInt64, UInt64}`.

devmotion: In this case you could still use a `UInt64` sampler but sample twice for each `subrng`. I guess you could dispatch on the type of subrng to simplify the code.

yebai (approved):

src/container.jl
```diff
@@ -190,12 +197,34 @@ function update_keys!(pc::ParticleContainer, ref::Union{Particle,Nothing}=nothin
...
+Set seeds of particle rng from user-provided `rng`
+"""
+function seed_from_rng!(
+    pc::ParticleContainer{T,<:TracedRNG{R,N,<:Random123.AbstractR123{I}}},
+    rng::Random.AbstractRNG,
+    ref::Union{Particle,Nothing}=nothing,
+) where {T,R,N,I}
+    n = length(pc.vals)
+    nseeds = isnothing(ref) ? n : n - 1
+    rngs = vcat([pc[i].rng for i in 1:nseeds], [pc.rng])
```
devmotion, line 218: This allocation could be avoided by splitting the loop.

FredericWantiez: Yes, I've changed it, we don't really need to collect these

PR #43, opened 2022-02-20
Switch to Libtask.TapedTask

rikhuijzer:
Similar to https://github.com/TuringLang/Turing.jl/pull/1778.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AdvancedPS"
 uuid = "576499cb-2369-40b2-a588-c64705576edc"
 authors = ["TuringLang"]
-version = "0.3.0"
+version = "0.3.4"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AdvancedPS"
 uuid = "576499cb-2369-40b2-a588-c64705576edc"
 authors = ["TuringLang"]
-version = "0.3.4"
+version = "0.4"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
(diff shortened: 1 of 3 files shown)


devmotion (commented):
Can we use `task` instead of `ttask`? Seems a bit simpler if we change the name.

rikhuijzer:
> Can we use `task` instead of `ttask`? Seems a bit simpler if we change the name.

Libtask uses `ttask` to refer to objects of type `TapedTask`. This object also contains a field `task::Task`, so that's why separate variable names are used.

devmotion:
Yeah but in `Trace` we don't have any other task. And the type doesn't seem very relevant from the perspective of AdvancedPS.

Project.toml
```diff
@@ -14,7 +14,7 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 [compat]
 AbstractMCMC = "2, 3"
 Distributions = "0.23, 0.24, 0.25"
-Libtask = "0.6.7"
+Libtask = "0.7"
```
devmotion, line 17: The compat shouldn't be changed:

```suggestion
Libtask = "0.6.7"
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "AdvancedPS"
 uuid = "576499cb-2369-40b2-a588-c64705576edc"
 authors = ["TuringLang"]
-version = "0.3.0"
+version = "0.3.4"
```
yebai, line 4:
```suggestion
version = "0.4"
```
the `master` branch contains breaking changes that have not been released yet. The most recent `v0.3x` releases are based on the [`releases` ](https://github.com/TuringLang/AdvancedPS.jl/tree/releases) branch.

yebai (approved):

rikhuijzer:
@yebai could you manage the release or  provide more details on how the release process is going here at AdvancedPS?

yebai:
> @yebai could you manage the release or provide more details on how the release process is going here at AdvancedPS?

Fixed by https://github.com/TuringLang/AdvancedPS.jl/pull/44

PR #45, opened 2022-02-21
Switch to Libtask 0.7

rikhuijzer:
I'll also make a PR in the releases branch.

Oh wait, Hong already set the Libtask to 0.7 on the 0.3.5 release. I'll close this

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,7 +14,7 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 [compat]
 AbstractMCMC = "2, 3"
 Distributions = "0.23, 0.24, 0.25"
-Libtask = "0.6.7"
+Libtask = "0.7"
 Random123 = "1.3"
 StatsFuns = "0.9"
 julia = "1.3"
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -9,6 +9,6 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 AbstractMCMC = "2, 3"
 Distributions = "0.24, 0.25"
-Libtask = "0.6"
+Libtask = "0.7"
 julia = "1.3"
 Random123 = "1.3"


devmotion:
Probably we should change the `Trace` struct, otherwise the task field has an abstract type. This requires a breaking release though and additional changes in places like `Vector{Trace}`.

devmotion:
>  Hong already set the Libtask to 0.7 on the 0.3.5 release.

We could fix this compat entry in the registry if we want to avoid the abstract fields.

rikhuijzer:
> We could fix this compat entry in the registry if we want to avoid the abstract fields.

I've now proposed to pin it at Turing (https://github.com/TuringLang/Turing.jl/pull/1779).

Maybe it would actually be good to move AdvancedPS#master to Libtask 0.7 too to avoid possible confusion.

Sorry for me being all over the place here. I'm having zero experience with a staging environment and am having trouble to understand how things go here in this repo.

devmotion:
My point is independent of the AdvancedPS branches: It was a mistake to declare any version of AdvancedPS to be compatible with Libtask 0.7. AdvancedPS has to be updated properly first, e.g., by changing the `Trace` struct - but since this involves breaking changes it has to go into a breaking release. So IMO there should not be any 0.3.x release that it is declared compatible with Libtask 0.7.

rikhuijzer:
> So IMO there should not be any 0.3.x release that it is declared compatible with Libtask 0.7.

Just to be sure. You mean "shouldn't have" because 0.3.5 has Libtask 0.7 in the compat entry (https://github.com/TuringLang/AdvancedPS.jl/blob/releases/Project.toml).

devmotion:
Yes. But that can be fixed in the registry.

yebai:
>You mean "shouldn't have" because 0.3.5 has Libtask 0.7 in the compat entry (https://github.com/TuringLang/AdvancedPS.jl/blob/releases/Project.toml).


It was motivated to fix the `Libtask` integration test. Since the `releases` branch is only meant for hotfixes, it might be ok to keep the compact entry there -- if it does not cause any harm. Otherwise, we will have to either 

1. make a new release from `AdvancedPS#master`, 
2. or, keep backporting new changes to `AdvancedPS@releases` once Turing requires `Libtask@0.7`

yebai (approved):

devmotion:
> it might be ok to keep the compact entry there -- if it does not cause any harm.

The problem is that it's not a simple update - suddenly all nicely typed structs and containers have abstract fields or type parameters. IMO it would have been cleaner to make a 0.4 release with the Libtask 0.7 compat update and fix all these issues at the same time. I guess we could make a 0.4 release on the releases branch if we don't want to release the other changes on the master branch right now, and then update the master branch to 0.5. Alternatively, if the releases branch should only contain 0.3.x releases we could fork another releases-0.4 branch from it and add the compat update there.

devmotion:
Yes, in #44 only `CTask` is replaced with the alias `TapedTask` (and the field `ctask` was renamed to `task` which caused a hiccup in Turing). This is completely fine, the alias existed in Libtask 0.6.7-0.6.x so the compat entries did not have to be updated.

In Libtask 0.7, however, `TapedTask` is not a concrete type anymore since a type parameter was added. This is potentially breaking since, as in AdvancedPS, structs with a field of type `TapedTask` that was concretely typed now contain a field with an abstract type `TapedTask`. Even more importantly, it breaks dispatches on e.g. `Vector{TapedTask}` since this won't catch `Vector{TapedTask{F}}` for some `F`. Since at least one of these issues shows up in AdvancedPS (not sure about dispatches, haven't checked the code in detail) updating the Libtask compatibility 0.7 can have undesired (and potentially breaking) consequences. Therefore I think updating to Libtask 0.7 should be done more carefully and after considering and fixing these issues.

Project.toml
```diff
@@ -14,7 +14,7 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 [compat]
 AbstractMCMC = "2, 3"
 Distributions = "0.23, 0.24, 0.25"
-Libtask = "0.6.7"
+Libtask = "0.7"
```
yebai, line 17:
```suggestion
Libtask = "0.7, 0.8"
```

test/Project.toml
```diff
@@ -9,6 +9,6 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 [compat]
 AbstractMCMC = "2, 3"
 Distributions = "0.24, 0.25"
-Libtask = "0.6"
+Libtask = "0.7"
```
yebai, line 12:
```suggestion
Libtask = "0.7, 0.8"
```

yebai (approved):

devmotion:
I see, if there are more changes to the type to be expected in the next weeks it might be easier to update AdvancedPS once everything has stabilized. Maybe we could open an issue that reminds us about these abstract fields.

In general, I think we shouldn't update compat bounds before versions of upstream packages are released. I don't think it makes updates much faster and it means if accidentally something breaks users will actually end up with the broken versions. Whereas if we update compat bounds once upstream packages are released, hopefully our tests catch all (common) bugs.

yebai:
I'm merging this PR, but we won't make a new release from `master` until #32 is complete. In addition, I believe some (minor) changes are required from the Turing side to become compatible with `AdvancedPS#master`.

PR #67, opened 2022-07-08
Update GenericModel for Turing evaluator

FredericWantiez:
Turing `evaluator` logic does not fit with the `GenericModel{F}` definition. We need something like `GenericModel{F1,F2}` for the `evaluator` and the `model` itself


FredericWantiez:
We king of need this one for the upstream to Turing to compile: https://github.com/TuringLang/Turing.jl/pull/1858

PR #69, opened 2022-10-29
Improving test coverage

FredericWantiez:
Given all the changes on `master` that happened recently it would be good to improve the coverage a bit, especially on the `PGAS` and `TracedRNG` parts.


FredericWantiez:
I plan to add more tests after this. This is mainly to clean up and remove `TArray` references

src/rng.jl
```diff
@@ -117,6 +117,3 @@ Random123.set_counter!(r::TracedRNG, n::Integer) = r.count = n
 Increase the model step counter by `n`
 """
 inc_counter!(r::TracedRNG, n::Integer=1) = r.count += n
-
-# Handle mutating RNG properly when forking the tape in Libtask
-Libtask.tape_copy(rng::TracedRNG) = copy(rng)
```
yebai, line 122: By default,`Libtask.TapedTask` only deep copies `Array` and `Ref`. See [here](https://github.com/TuringLang/Libtask.jl/blob/master/src/tapedtask.jl#L68) for details. So, we might need to add `TracedRNG` to the `TapedTask` arguments to enable customised deep copying.

yebai: In addition, we can override the deep and shallow copy behaviour of Julia types by overloading the `tape_shallowcopy` and `tape_deepcopy` functions. See [here](https://github.com/TuringLang/Libtask.jl/blob/master/src/tapedfunction.jl#L456).

yebai:
thanks, @FredericWantiez - I'll merge this PR so we can finalise the Turing update. Please feel free to open another PR for more tests.

Issue #77, opened 2023-08-24
PGAS errors out when adpting for nonlinear time series example


YSanchezAraujo:
Working through [this](https://github.com/TuringLang/AdvancedPS.jl/blob/master/examples/particle-gibbs/script.jl) example, and it suggests using PGAS to help with sample impoverishment. I've tried only replacing `pgas = AdvancedPS.PG(n_particles)` with `pgas = AdvancedPS.PGAS(n_particles)` as is done [here](https://github.com/TuringLang/AdvancedPS.jl/blob/master/examples/gaussian-ssm/script.jl) , while keeping all else the same but am met with this error

```
ERROR: MethodError: no method matching step(::MersenneTwister, ::NonLinearTimeSeries, ::AdvancedPS.PGAS{AdvancedPS.ResampleWithESSThreshold{typeof(resample_systematic), Float64}})
Closest candidates are:
  step(::AbstractRNG, ::AdvancedPS.AbstractStateSpaceModel, ::AdvancedPS.PGAS) at ~/.julia/packages/AdvancedPS/Vox9w/src/smc.jl:130
  step(::AbstractRNG, ::AdvancedPS.AbstractStateSpaceModel, ::AdvancedPS.PGAS, ::Union{Nothing, AdvancedPS.PGState}; kwargs...) at ~/.julia/packages/AdvancedPS/Vox9w/src/smc.jl:130
  step(::AbstractRNG, ::AbstractMCMC.AbstractModel, ::AdvancedPS.PG) at ~/.julia/packages/AdvancedPS/Vox9w/src/smc.jl:88
  ...
Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/AbstractMCMC/fWWW0/src/sample.jl:125 [inlined]
  [2] macro expansion
    @ ~/.julia/packages/ProgressLogging/6KXlp/src/ProgressLogging.jl:328 [inlined]
  [3] (::AbstractMCMC.var"#21#22"{Bool, String, Nothing, Int64, Int64, Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, MersenneTwister, NonLinearTimeSeries, AdvancedPS.PGAS{AdvancedPS.ResampleWithESSThreshold{typeof(resample_systematic), Float64}}, Int64, Int64})()
    @ AbstractMCMC ~/.julia/packages/AbstractMCMC/fWWW0/src/logging.jl:12
[...]

PR #79, opened 2023-09-04
Update script.jl

mattiasvillani:
Fixed some typos in the model and one in the code (variance was used instead of standard deviation in measurement model). Added PGAS sampling at the end to show that it solves the degeneracy problem, which should close the issue https://github.com/TuringLang/AdvancedPS.jl/issues/77

Please check my text 'To use this sampler we need to define the transition and observation densities as well as the initial distribution in the following way:' as I am not sure that this is needed only for PGAS (since it was not needed for PG).

### DIFF
--- a/examples/particle-gibbs/script.jl
+++ b/examples/particle-gibbs/script.jl
@@ -1,4 +1,3 @@
-# # Particle Gibbs for non-linear models
 using AdvancedPS
 using Random
 using Distributions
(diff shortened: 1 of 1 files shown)


examples/particle-gibbs/script.jl
```diff
@@ -88,8 +86,8 @@ end
 
 # Here we use the particle gibbs kernel without adaptive resampling.
 model = NonLinearTimeSeries(θ₀)
-pgas = AdvancedPS.PG(Nₚ, 1.0)
-chains = sample(rng, model, pgas, Nₛ; progress=false);
```
FredericWantiez, line 92: Minor comment, but why removing the `rng` argument ? That makes the doc reproducible (to some extent, rng implementations can change with julia versions)

FredericWantiez:
The `PGAS` part looks correct to me, we only need to define the densities for `PGAS` since the sampler targets models with a special structure (SSM). `PG` uses `Libtask` in the background to allow for somewhat arbitrary models (through program tracing).

examples/particle-gibbs/script.jl
mattiasvillani: sorry, removed that when trying to get pgas to work. I edited the line now so that it is back to using rng as first argument, and also progress=false as in the original example.

YSanchezAraujo:
> The `PGAS` part looks correct to me, we only need to define the densities for `PGAS` since the sampler targets models with a special structure (SSM). `PG` uses `Libtask` in the background to allow for somewhat arbitrary models (through program tracing).

Does that mean that more complicated / arbitrary models aren't compatible with `PGAS` ? Is a work around possible ? Happy to look into it if pointed in the right direction

mattiasvillani:
My understanding is the essentially arbitrary models are possible (I am currently working with a non-standard one and it works), but that you need to define the transition and observation densities as well as the initial distribution as done in my PR.

YSanchezAraujo:
> My understanding is the essentially arbitrary models are possible (I am currently working with a non-standard one and it works), but that you need to define the transition and observation densities as well as the initial distribution as done in my PR.

I think this is where some confusion would come in. The way it is shown here, those functions define some distribution that is used, but don't specify how they are used. In the old example, they are explicitly used along with `rng` to generate samples. This made it easy to reason about what one does when the transition (for example) involves sampling from multiple distributions, or consists of dependent steps. But less clear as to what one does in this case.

mattiasvillani:
Note that the old example (using PG) is still there, I just added PGAS as a follow-up (as the old example ended rather disappointingly with degeneracy and an Issue asked for how to use PGAS here). 
I think it would be nice to explain how these distributions are used (it is cloudy for me at the moment, I am now trying to get more into the code), but I note that the essentially the same (lack of) explanation is in the other PGAS example: https://turinglang.org/AdvancedPS.jl/dev/examples/gaussian-ssm/.

FredericWantiez:
The documentation is rather sparse I agree, thanks for the feedback I'll spend some time improving it. 

For the model part. We've only implemented `PGAS` for State Space Models since the ancestor sampling scheme is  tractable for these [[from the original article, eq. 17]](https://jmlr.org/papers/volume15/lindsten14a/lindsten14a.pdf):
image
 where  $f\_{\theta}$ corresponds to the `transition` function in the code. 
The `initialization` and `observation` functions are used to update the particles and compute the importance weights.

For `PG`, we don't need to impose any particular structure as long as we can store/regenerate the trajectories of the reference particle. That's why you can use `PG` for arbitrary models, you just need to emit the `logpdf` from the `model`, something like this in the code. 
```julia
function (model::NonStandardModel)(rng::Random.AbstractRNG)
    logpdf = simulation_steps(rng, ...)
    Libtask.produce(logpdf)
end
```

Hopefully that helps. If you have an example that's causing confusion, happy to look into it.

mattiasvillani:
Oh, I see. The question was about models outside the state-space class, non-Markovian models. Thanks for the clarification. That restriction of PGAS should be mentioned in the docs where PGAS is introduced. Let me know if I should change anything here.

examples/particle-gibbs/script.jl
```diff
@@ -137,3 +134,38 @@ plot(
...
+    return logpdf(g(model, state, step), y[step])
+end
+AdvancedPS.isdone(::NonLinearTimeSeries, step) = step > Tₘ
+
+# We can now sample from the model using the PGAS sampler and collect the trajectories.
+pg = AdvancedPS.PGAS(Nₚ)
+chains = sample(model, pg, Nₛ);
+trajectories = map(chains) do sample
+    replay(sample.trajectory)
+end
```
FredericWantiez, line 152: This is already done earlier in the code and `PGAS` stores the individual trajectories, no need to replay 
```suggestion
```

FredericWantiez:
A very small thing, but the rest looks fine to me. Thanks again for looking into this !

PR #80, opened 2023-09-18
Fixed 

mattiasvillani:
The changes in PR #79 led to the following bugs (some edits in #79 following the review strangely got lost in the final version):

- The title got removed from the example doc
- Long printout of not important vector
- The replay of the trajectories somehow got killed in edits following the PR preview which led the pgas trajectories not showing in the plot.
- The rng argument that was requested to be added to the pgas call also got wiped out in the edits from the review.

This PR fixes all four issues.

### DIFF
--- a/examples/particle-gibbs/script.jl
+++ b/examples/particle-gibbs/script.jl
@@ -1,3 +1,4 @@
+# # Particle Gibbs for non-linear models
 using AdvancedPS
 using Random
 using Distributions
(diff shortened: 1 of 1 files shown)


examples/particle-gibbs/script.jl
```diff
@@ -145,10 +146,13 @@ end
...
-pg = AdvancedPS.PGAS(Nₚ)
-chains = sample(model, pg, Nₛ);
-particles = hcat([trajectory.model.f.X for trajectory in trajectories]...)
-mean_trajectory = mean(particles; dims=2)
+pgas = AdvancedPS.PGAS(Nₚ)
+chains = sample(rng, model, pgas, Nₛ; progress=false);
+trajectories = map(chains) do sample
+    replay(sample.trajectory)
+end
+particles = hcat([trajectory.model.f.X for trajectory in trajectories]...);
```
FredericWantiez, line 154: This should have been (without using the suggestion tool, since I messed up everything last time):
```julia
particles = hcat([trajectory.model.f.X for trajectory in chains]...); # pulling it from the chains directly
```
and you should be able to remove the `replay` part

mattiasvillani: I would need to read up on the replay idea here, but replacing 

`trajectories = map(chains) do sample
    replay(sample.trajectory)
end
particles = hcat([trajectory.model.f.X for trajectory in trajectories]...);`

with 

`particles = hcat([trajectory.model.f.X for trajectory in chains]...); # pulling it from the chains directly
`

gives me the error:

ERROR: type PGSample has no field model

FredericWantiez: This fixes this issue
```julia
particles = hcat([chain.trajectory.model.X for chain in pgas_chains]...)
```
but seems to hit another one with the way we copy models around. I'll open an issue and look into it

FredericWantiez: Agree, since the fix pretty much requires https://github.com/TuringLang/AdvancedPS.jl/pull/75 to be merged first

FredericWantiez (approved):

PR #91, opened 2023-11-04
Proper type dispatch

FredericWantiez:


FredericWantiez:
You're so fast ! Thanks for the review

PR #93, opened 2023-11-10
Update AdvancedPS to match SSMProblems interface

THargreaves:
Draft work on updating AdvancedPS to match the SSMProblems interface.

To run this code you need to `dev` SSMProblems in the AdvancedPS environment and `dev` AdvancedPS in the examples/particle-gibbs environment.

### DIFF
--- a/examples/particle-gibbs/script.jl
+++ b/examples/particle-gibbs/script.jl
@@ -1,5 +1,6 @@
 # # Particle Gibbs for non-linear models
 using AdvancedPS
+using SSMProblems
 using Random
 using Distributions
 using Plots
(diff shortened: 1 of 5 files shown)


examples/particle-gibbs/script.jl
```diff
@@ -89,25 +91,25 @@ plot(x; label="x", xlabel="t")
...
+)
+    return logpdf(g(state, model), observation)
 end
-AdvancedPS.isdone(::NonLinearTimeSeries, step) = step > Tₘ
 
-# Here we use the particle gibbs kernel without adaptive resampling.
-model = NonLinearTimeSeries(θ₀)
 pg = AdvancedPS.PG(Nₚ, 1.0)
-chains = sample(rng, model, pg, Nₛ; progress=false);
+chains = sample(rng, model, pg, Nₛ; observations=y, progress=false);
```
FredericWantiez, line 112: The main issue here is that `AbstractMCMC` does not forward data explicitly, it just assumes the obervables are internal to the model

src/container.jl
```diff
@@ -256,20 +256,20 @@ end
...
+    for i in eachindex(particles)
         p = particles[i]
 
         # Obtain ``\\log p(yₜ | y₁, …, yₜ₋₁, x₁, …, xₜ, θ₁, …, θₜ)``, or `nothing` if the
         # the execution of the model is finished.
         # Here ``yᵢ`` are observations, ``xᵢ`` variables of the particle filter, and
         # ``θᵢ`` are variables of other samplers.
         isref = p === ref
-        score = advance!(p, isref)
+        score = advance!(p, y, isref)
```
FredericWantiez, line 272: That would make backward compatibility with `Libtask` quite difficult

examples/particle-gibbs/script.jl
THargreaves: I see. One way around this would be to define something like

```
struct ConditionedSSM
    ssm::AbstractStateSpaceModel
    observations::Vector{Vector{Float64}}
end
```

though that's one more bit of API bloat.

FredericWantiez: Doesn't have to be nested, or another field.
```julia
function emission_logensity(model::AbstractStateSpaceModel, step::Int) 
   return logpdf(Normal(), observations[step]) # Global scope
end
```
or 
```julia
function emission_logensity(model::AbstractStateSpaceModel, step::Int) 
   return logpdf(Normal(), model.observations[step]) 
end
```

It's a bit what `model() | (; y=observations)` would do in `DynamicPPL`

PR #94, opened 2023-11-30
Levy SSM inference using Particle Gibbs

THargreaves:
An example implementation of the Levy State Space Model (Godsill et al. 2020). Simulation of the driving Levy process is performed using the generalised shot noise formulation.

### DIFF
--- a/examples/levy-ssm/script.jl
+++ b/examples/levy-ssm/script.jl
@@ -1,0 +1,20 @@
+# # The Levy State Space Model (Godsill et al. 2020)
+
+include("simulation.jl")
+
+Parameters = @NamedTuple begin
+    μ_W::Float64  # Subordinator process mean
+    σ_W::Float64  # Subordinator process variance
+    β::Float64    # Tempering parameter
+    C::Float64    # Jump density
+end
+
+# Simulation parameters
+T = 100    # Time horizon
+res = 100  # Sampling resolution
+
+# Sampling parameters
+
+
+
+
--- a/examples/levy-ssm/simulation.jl
+++ b/examples/levy-ssm/simulation.jl
@@ -1,0 +1,121 @@
+# # Code for simulating Levy processes
+using Distributions
+using Plots
+
+abstract type LevyPointProcess end
+
+# TODO: check naming conventions
+function integrate(::LevyPointProcess, ts, jumps, jump_times)
+    # TODO: are these already sorted
+    samples = []
+    for t in ts
+        subset = jumps[jump_times .<= t]
+        # Sum of empty array has no clear type
+        if length(subset) > 0
+            push!(samples, sum(subset))
+        else
+            push!(samples, 0.0)
+        end
+    end
+    return samples
+end
+
+struct NormalGammaProcess <: LevyPointProcess
+    μ_W::Float64  # Subordinator process mean
+    σ_W::Float64  # Subordinator process variance
+    β::Float64    # Tempering parameter
+    C::Float64    # Jump density
+    T::Float64    # Time horizon
+end
+
+function simulate(ngp::NormalGammaProcess, res::Int)
+    ts = collect(range(0, ngp.T, length=res))
+    # TODO: check naming against function internals
+    gamma_paths, subordinator_jumps, jump_times = simulate_gamma_process(ngp, ts)
+
+    NVM_jumps = (
+        ngp.μ_W * subordinator_jumps +
+        ngp.σ_W * sqrt.(subordinator_jumps) .* randn(length(subordinator_jumps))
+    )
+
+    samples = integrate(ngp, ts, NVM_jumps, jump_times)
+    return samples, NVM_jumps, jump_times, subordinator_jumps
+end
+
+function simulate_gamma_process(ngp::NormalGammaProcess, ts::Vector{Float64})
+    # Generate Poisson epochs over [0, T]
+    # TODO: benchmark this vs. uniform sampling conditioned on N ~ Pois(T)
+    poisson_epochs = []
+    t = 0  # current time
+    expired = false
+    while !expired
+        t += rand(Exponential(1 / ngp.T))
+        if t < ngp.T
+            push!(poisson_epochs, t)
+        else
+            expired = true
+        end
+    end
+
+    # Generate samples
+    Xs = []
+    for t in poisson_epochs
+        X = 1 / (ngp.β * (exp(t / ngp.C) - 1))
+        p = (1 + ngp.β * X) * exp(-ngp.β * X)
+        if rand() < p
+            push!(Xs, X)
+        end
+    end
+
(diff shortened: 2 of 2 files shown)


examples/levy-ssm/simulation.jl
```diff
@@ -0,0 +1,84 @@
...
+end
+
+function simulate_gamma_process(ngp::NormalGammaProcess, ts::Vector{Float64})
+    # Generate Poisson epochs over [0, T]
+    # TODO: benchmark this vs. uniform sampling conditioned on N ~ Pois(T)
+    poisson_epochs = Float64[]
+    t = 0  # current time
+    expired = false
+    while !expired
+        t += rand(Exponential(1 / ngp.T))
```
FredericWantiez, line 41: Shouldn't that be `dt` instead ?

THargreaves: I'm not too sure. I actually can't remember where I got this from.

In Algo 2 of [Godsill, 2022](https://arxiv.org/pdf/2105.09429.pdf), we have that the Poisson epochs are unit rate. 

It looks like what I've done is taken these and stretched these to [0, T]. (Note, Exponential uses scale parameterisation so the rate parameter here is T).

I think this might have been me confusing the Poisson process used to generate the jump intensities and the Poisson process that generates the jump times.

What makes you think `1 / dt`?

THargreaves: I think you might be right. Yaman's code has 

```
t_series, x_series = self.model.I.subordinator.simulate_points(rate=dt, low=times[i-1], high=times[i], n_particles=1)
```

With this rate then passed through to the Poisson rate.

examples/levy-ssm/gamma_process.jl
```diff
@@ -0,0 +1,176 @@
...
+        truncated = last_jump < truncation
+        while !truncated
+            t += rand(Exponential(1.0 / rate))
+            xi = 1.0 / (beta * (exp(t) / C - 1))
+            prob = (1.0 + beta * xi) * exp(-beta * xi)
+            if rand() < prob
+                push!(jumps, xi)
+                last_jump = xi
+            end
+            truncated = last_jump < truncation
```
THargreaves, line 52: What paper is this step coming from. To my understanding, in the original LevySSM paper we truncate the series when `t` (the Poisson epochs) exceed some truncation parameter `c` multiplied by `dt`.

examples/levy-ssm/simulation.jl
FredericWantiez: Mostly from this [paper](https://arxiv.org/pdf/2309.11422.pdf) where everything is defined in (s, t) 

in your 1D case, the jump generation process ends up cycling a very large number of times

examples/levy-ssm/gamma_process.jl
FredericWantiez: I think Yaman uses a more fancy adaptive truncation scheme

examples/levy-ssm/gamma_process.jl
```diff
@@ -0,0 +1,176 @@
...
+    let beta = process.beta, C = process.C
+        jumps = Float64[]
+        last_jump = Inf
+        t = t0
+        truncated = last_jump < truncation
+        while !truncated
+            t += rand(Exponential(1.0 / rate))
+            xi = 1.0 / (beta * (exp(t) / C - 1))
+            prob = (1.0 + beta * xi) * exp(-beta * xi)
+            if rand() < prob
```
FredericWantiez, line 48: Should be `if rand() < prob && xi >= truncation`

THargreaves: xi is monotonic in t so it makes more sense to check if xi < truncation immediately after generating it and then breaking then. Just need to be careful that that can then lead to an empty jump array.

Otherwise I think you actually get stuck in an infinite loop since last_jump is never updated.

THargreaves:
```
while true
     t += rand(Exponential(1.0 / rate))
     xi = 1.0 / (beta * (exp(t) / C - 1))
     if xi < truncation
         break
     end
     prob = (1.0 + beta * xi) * exp(-beta * xi)
     if rand() < prob
         push!(jumps, xi)
     end
```

examples/levy-ssm/script.jl
```diff
@@ -0,0 +1,257 @@
...
+
+function Base.rand(rng::Random.AbstractRNG, td::TransitionDistribution)
+    let model = td.model, s = td.s, t = td.t, state = td.current_state
+        dt = t - s
+        path = simulate(model.θ.process, dt, s, t, ϵ)
+        μ, Σ = meancov(t, model.θ.dyn, path, model.θ.nvm)
+        Σ += 1e-6 * I
+        return MixedState(
+            rand(rng, MultivariateNormal(exp(dyn, dt) * state.x + μ, Σ)), path
+        )
```
FredericWantiez, line 176: Can't we just mutate the state instead ?

THargreaves: That works in a purely filtering context, but if you then want to apply smoothing, you need all of the historic states saved, so mutating isn't an option.

I think this clunkiness is best solved by modifying AdvancedPS.jl to expect a sample to be returned rather than a distribution. I only wrote it this way to maintain compatibility.

THargreaves:
I think this is as far as we can go with this commit with making substantial changes to AdvancedPS.jl — specifically, changing AdvancedPS.jl to expect a sample to be return by the transition function rather than a distribution. 

My feeling is that this is at a good point to merge. @yebai @FredericWantiez?

examples/levy-ssm/script.jl
```diff
@@ -0,0 +1,275 @@
...
+end
+
+# Required for ancestor sampling in PGAS
+function Distributions.logpdf(td::TransitionDistribution, state::MixedState)
+    let model = td.model, s = td.s, t = td.t, state = td.current_state
+        dt = t - s
+        path = simulate(model.θ.process, dt, s, t, ϵ)
+        μ, Σ = meancov(t, model.θ.dyn, path, model.θ.nvm)
+        Σ += 1e-6 * I
+        return logpdf(MultivariateNormal(exp(dyn, dt) * state.x + μ, Σ), state.x)
```
FredericWantiez, line 202: We're not using `state` ? Looks like should be something like:
```julia
let model = td.model, s = td.s, t = td.t, prev_state = td.current_state
  ...
  return logpdf(MultivariateNormal(exp(dyn, dt) * prev_state.x + μ, Σ), state.x)
end
```

FredericWantiez:
Apologies for making it a mess, but that's how it would look like with `SSMProblems`: [here](https://github.com/TuringLang/AdvancedPS.jl/pull/97/files#diff-820583ee6a88859c70697091ef200557c5d44fae07731c40779027029a508b96)

PR #97, opened 2024-03-24
SSMProblems integration

FredericWantiez:
The two main issues I see with the current API in `SSMProblems`:
- Handling of observations, I wrapped them up in the model and the user has to deal with it. 
- Couldn't easily remove `AdvancedPS.isdone`


THargreaves:
This looks good to me. Happy for this to replace the previous partial PRs.

Wrapping the observations in the model seems like a sensible approach for now.

FredericWantiez:
@yebai

PR #109, opened 2025-01-21
Make particle Gibbs deterministic across architectures

### LINKED ISSUE #108: Particle Gibbs reproducibly gives different results on x64 vs x86
Not for the first time, x86 CI gives us mysterious problems, in that some numerical tests fail on x86 despite all RNGs being thoroughly seeded (i.e. the results **should** be fully deterministic).

For example, using this `runtests.jl` on x64 and x86 GitHub runners will give different results at the end, see https://github.com/penelopeysm/Shaymin.jl/actions/runs/12412993388/job/34653933016

```julia
using Test
using StableRNGs
using Random
using Turing

Random.seed!(468)

@testset verbose = true "Shaymin.jl" begin
    @testset "using global seed" begin
        x1 = randn(3)
        @info x1
    end

    @testset "stablerng" begin
        x2 = randn(StableRNG(468), 3)
        @info x2
    end

    @testset "pg"  begin
        @model function f(y)
            a ~ Normal(0, 1)
            y ~ Normal(a, 1)
        end
        Random.seed!(468)
        alg = PG(15)
        chain = sample(StableRNG(468), f(1.5), alg, 50; progress=false)
        @show mean(chain[:a])
    end
end
```

```julia
# x64
[ Info: [0.07200886749732076, -0.0740437565595174, 0.6327762377562545]
[ Info: [1.2876157288026433, -0.2953479054222536, -1.205615981210787]
mean(chain[:a]) = 0.7475257036106626

# x86
[ Info: [0.07200886749732076, -0.0740437565595174, 0.6327762377562545]
[ Info: [1.2876157288026433, -0.2953479054222536, -1.205615981210787]
mean(chain[:a]) = 0.7973086809553678
```

Note that:

1. The results are deterministic if run repeatedly **on the same architecture**, so the problem isn't that the implementation doesn't use the provided `rng`;
2. The first two testsets with `Random.randn(10)` and `rand(StableRNG(468), 10)` are deterministic across architectures, so it's not a mistake in the implementation of the random number generator.

penelopeysm:
Fixes #108 

The problem was in this function `split`:

https://github.com/TuringLang/AdvancedPS.jl/blob/379336b8fcd662baa4a227100c94aefe491d62fe/src/rng.jl#L33-L41

Here, `UInt(1)` gives `UInt64` on 64-bit systems., and `UInt32` on 32-bit. Thus, the result of `hash` was different, despite ensuring that we convert back to the same type at the end.

The 'simple' solution of replacing `UInt` with `UInt64` doesn't work, because on 32-bit systems, `hash(::UInt64, ::UInt64)` is not defined.

This PR therefore uses a pseudo-random number generator to deterministically generate new values of type `T` from the original value.

### DIFF
--- a/test/linear-gaussian.jl
+++ b/test/linear-gaussian.jl
@@ -26,7 +26,7 @@ end
 @testset "linear-gaussian.jl" begin
     T = 3
     N_PARTICLES = 100
-    N_SAMPLES = 50
+    N_SAMPLES = 200
 
     # Model dynamics
     a = 0.5
@@ -100,12 +100,14 @@ end
     @testset "PGAS" begin
         pgas = AdvancedPS.PGAS(N_PARTICLES)
         p = test_algorithm(rng, pgas, model, N_SAMPLES, Xf)
+        @info p
         @test p > 0.05
     end
 
     @testset "PG" begin
         pg = AdvancedPS.PG(N_PARTICLES)
         p = test_algorithm(rng, pg, model, N_SAMPLES, Xf)
+        @info p
         @test p > 0.05
     end
 end
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AdvancedPS"
 uuid = "576499cb-2369-40b2-a588-c64705576edc"
 authors = ["TuringLang"]
-version = "0.6"
+version = "0.6.1"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/src/rng.jl
+++ b/src/rng.jl
@@ -36,8 +36,9 @@ Base.rand(rng::TracedRNG, ::Type{T}) where {T} = Base.rand(rng.rng, T)
 Split `key` into `n` new keys
 """
 function split(key::Integer, n::Integer=1)
-    T = typeof(key) # Make sure the type of `key` is consistent on W32 and W64 systems.
-    return T[hash(key, i) for i in UInt(1):UInt(n)]
+    T = typeof(key)
+    inner_rng = Random.MersenneTwister(key)
+    return rand(inner_rng, T, n)
 end
 
 """


test/linear-gaussian.jl
```diff
@@ -26,7 +26,7 @@ end
 @testset "linear-gaussian.jl" begin
     T = 3
     N_PARTICLES = 100
-    N_SAMPLES = 50
+    N_SAMPLES = 150
```
penelopeysm, line 29: I had to bump this to get the test to pass.

test/linear-gaussian.jl
```diff
@@ -26,7 +26,7 @@ end
 @testset "linear-gaussian.jl" begin
     T = 3
     N_PARTICLES = 100
-    N_SAMPLES = 50
+    N_SAMPLES = 200
```
penelopeysm, line 29: I had to bump this to get the test to pass.

penelopeysm:
It's not possible to put in a test that checks that it's the same result across architectures (we could compare to a fixed value, but that would be very brittle). But this at least demonstrates that sampling a Turing model with `PG()` now gives the same results across architectures: https://github.com/penelopeysm/Shaymin.jl/actions/runs/12891706831

mhauru (approved):
Great work figuring this out! This has been bugging me, too, for a while now.

penelopeysm:
Thank you, @yebai, I was reading this over lunch :)

For context, the RNG stream being used right now is Philox which is described here: https://www.thesalmons.org/john/random123/papers/random123sc11.pdf

I'm not really certain that our original implementation of `split` is correct either, though - is it? It seems to me that (unlike, e.g., Numpy, https://numpy.org/doc/2.1/reference/random/bit_generators/philox.html) the Random123.jl package doesn't include the necessary functionality to split a Philox RNG, and this split function seems to me to be quite ad hoc too as the algorithm isn't mentioned in the paper.

FredericWantiez:
Thanks @penelopeysm for the fix. I don't think there was an implementation of `split` at the time, probably best to switch to `SplittableRandoms.jl` now

penelopeysm:
I'm in favour of switching over, and I think that it does have exactly the right functionality, but looking at AdvancedPS right now, it seems to me that that would require a fairly substantial reworking of the whole `rng.jl` file.

I'd prefer to merge this and open another issue for that (unless there are concerns over this PR itself, of course).

FredericWantiez:
We could also tweak `split` here to be somewhat closer to the jax implementation for example, but that's probably going to abuse the `Random123` interface. 
https://github.com/jax-ml/jax/blob/main/jax/_src/prng.py#L1114

PR #111, opened 2025-01-23
Update CI setup

penelopeysm:
This is broadly in line with what Turing / DynamicPPL has.

Main changes are:

- Bumping versions of actions
- Also changing 1.6 -> 'min' so that we don't have to update that in the future
- Also changing the 'nightly' test (brittle) to a new prerelease workflow
 - Using aarch64 on macOS rather than x64

While doing this I found that the minimum Julia bound of 1.6 was not correct: AdvancedPS depends on Libtask 0.8, which sets a minimum bound of Julia 1.7, so I updated that too.


.github/workflows/ci.yml
```diff
@@ -6,58 +6,69 @@ on:
...
+          version: ${{ matrix.runner.version }}
+          arch: ${{ matrix.runner.arch }}
+      - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
-        if: matrix.coverage
-      - uses: codecov/codecov-action@v1
-        if: matrix.coverage
+      - uses: codecov/codecov-action@v5
```
mhauru, line 67: Do we intend to run coverage stuff for all the matrix elements?

penelopeysm: I don't see why not, after all running coverage only on Ubuntu means that this will happen

```julia
if windows
   blah # coverage will claim you didn't test this line
else
   blah
end
```

penelopeysm: On DPPL/Turing coverage is run on all matrix combinations, too

mhauru: It seems DPPL does indeed, Turing does this instead: https://github.com/TuringLang/Turing.jl/blob/8bf98e1f29deb9e115f53ba8dccef590e3e24ce0/.github/workflows/Tests.yml#L92

I'm not opposed, as long as codecov knows how to aggregate the results reasonably and it doesn't slow the tests down too much. Just wanted to check.

mhauru (approved):
Seems like solid improvements all around

.github/workflows/ci.yml
penelopeysm: Oh, darn, OK. Thanks for the correction! I guess for AdvancedPS it's fine because the tests are so short, but I should probably quickly check whether there's a significant slowdown for DPPL.

PR #116, opened 2025-04-24
Update to most recent SSMProblems interface

charlesknipp:
This fork contains all of the required changes to support the newest SSMProblems interface. I also took the liberty to add some syntactical sugar (see `TracedSSM`), which I'm not entirely happy with.

- [x] unit tests pass
- [ ] examples run
  - [x] linear Gaussian model
  - [x] stochastic volatility
  - [ ] Gaussian process model (broken, but close)
  - [ ]  Levy process model (haven't touched)


charlesknipp:
I should note that this requires `dev SSMProblems` since we have not yet added `v0.5.2` to the registry, thus why CI is failing.

yebai:
https://github.com/JuliaRegistries/General/pull/129392 is blocking us from making a new `SSMProblems` release. Hopefully, this will get resolved in the next few days.

src/model.jl
```diff
@@ -10,6 +10,32 @@ const Particle = Trace
...
+    function TracedSSM(model::SSMProblems.StateSpaceModel{T,LD,OP}) where {T,LD,OP}
+        XT = eltype(LD)
+        YT = eltype(OP)
+        return new{SSMProblems.StateSpaceModel{T,LD,OP},XT,YT}(model, Vector{YT}())
+    end
+end
+
+(model::SSMProblems.StateSpaceModel)(Y::AbstractVector) = TracedSSM(model, Y)
+dynamics(ssm::TracedSSM, step::Int) = ssm.model.dyn
+observation(ssm::TracedSSM, step::Int) = ssm.model.obs
```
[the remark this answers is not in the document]
FredericWantiez: Looking at what I've done for the GP script, not sure it's correct given the non-markovian setup and the PGAS update implemented here

src/model.jl
```diff
@@ -10,6 +10,32 @@ const Particle = Trace
...
+    end
+
+    function TracedSSM(model::SSMProblems.StateSpaceModel{T,LD,OP}) where {T,LD,OP}
+        XT = eltype(LD)
+        YT = eltype(OP)
+        return new{SSMProblems.StateSpaceModel{T,LD,OP},XT,YT}(model, Vector{YT}())
+    end
+end
+
+(model::SSMProblems.StateSpaceModel)(Y::AbstractVector) = TracedSSM(model, Y)
```
FredericWantiez, line 33: A bit of type piracy, can't we just use a `TracedSSM` as the public API ?

```julia
model = TracedSSM(LinearGaussianStateSpaceModel(θ), Y)

sample(rng, traced_model)
```

yebai:
> https://github.com/JuliaRegistries/General/pull/129392 is blocking us from making a new SSMProblems release. 

This is now fixed!

src/model.jl
charlesknipp: Totally fine with me. I just figured that if we interface `SSMProblems` with `AbstractMCMC`, then we would have to define the observables and state space model tangentially. The syntactical sugar just gets around having to export something that doesn't necessarily need to be public.

charlesknipp:
## Review Comments

The Gaussian process model is something I struggled to replicate under the new system. Personally, I believe that non-Markovian models should not be handled by the SSMProblems API. With that being said, I think the GP example could be written in such a way that preserves Markov-ness. Any ideas are welcome.

I should note that the existence of `TracedSSM` seems redundant. It exists because we seek to create a `SSMTrace` with a `StateSpaceModel` object, which does not store the trajectories. Since none of the `Trace` objects are public facing, I abused some type piracy to preserve this feature. Which, as @FredericWantiez pointed out, is bad.

## Misc

I added some regards to the description. Contextualizing state space models in the Turing ecosystem really forced me to consider the way we handle sampling via `AbstractMCMC`. It's part of a much larger discussion, so I don't expect resolution in this PR.

yebai:
> With that being said, I think the GP example could be written in such a way that preserves Markov-ness. Any ideas are welcome.

Sparse GP priors would resolve this, since the GP prior over latent dynamics depend on a set of pseudo datapoints (known as inducing points) instead of the latent states (see Sec 4 of https://arxiv.org/abs/1306.2861). Then these pseudo datapoints are optimised conditioned on realisations of latent states (e.g. sampled trajectories from PMCMC).

yebai:
Thanks @charlesknipp!

PR #118, opened 2025-08-18
Allow passing kwargs on to Libtask (when it's an `AbstractTuringLibtaskModel`...)

penelopeysm:
This PR modifies LibtaskExt to allow sampling from Turing models with keyword arguments with SMC or PG. See https://github.com/TuringLang/Turing.jl/pull/2660 for usage.

It's obviously quite ugly, the problem is that even though Turing is the only library that subtypes the abstract type, the abstract type has to be defined in AdvancedPS so that the LibtaskExt can dispatch on it. Ideally, that abstract type + any relevant methods inside LibtaskExt should just be shifted into Turing proper. That's for another time.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AdvancedPS"
 uuid = "576499cb-2369-40b2-a588-c64705576edc"
 authors = ["TuringLang"]
-version = "0.7.1"
+version = "0.7.2"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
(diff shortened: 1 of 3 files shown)


yebai:
Yes, we'd like to merge the AdvancedPS code to Turing.jl. Libtask was holding us back; now, it is ready.

penelopeysm:
Hi @yebai, sorry just tidying this up before I go away :)

I am happy to try to work on upstreaming the necessary parts to Turing next year, but that might take more time, and this PR will unblock keyword arguments on SMC/PG on Turing (https://github.com/TuringLang/Turing.jl/pull/2660). So I think it makes sense to merge this first. I tested locally and it all works as intended, so happy for it to be reviewed now (cc also @mhauru).

sunxd3 (approved):
The logic looks good.

I am a bit worrying about introducing a struct with "Turing" in the naming (AbstractTuringLibtaskModel) into the namespace. This is purely aesthetic, of course. That said, not a blocker at all since we agree to move these into Turing pretty soon.

Thanks for the work, Penny!

PR #119, opened 2025-08-21
Update `SSMProblems` interface

charlesknipp:
With the addition of the state prior, there are a few minor changes that need to be made to the interface.

- [x] Update TracedSSM (would like some feedback here)
- [x] Fix unit test
- [x] Fix examples

I am not entirely happy about how I handle the `TracedSSM` containers since we have no type information without sampling from the state prior first.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,7 @@ Random = "<0.0.1, 1"
 Random123 = "1.3"
 Requires = "1.0"
 StatsFuns = "0.9, 1"
-SSMProblems = "0.5"
+SSMProblems = "0.6"
 julia = "1.10.8"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AdvancedPS"
 uuid = "576499cb-2369-40b2-a588-c64705576edc"
 authors = ["TuringLang"]
-version = "0.7"
+version = "0.7.1"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,6 +2,3 @@
 AdvancedPS = "576499cb-2369-40b2-a588-c64705576edc"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
-
-[compat]
-Documenter = "0.27"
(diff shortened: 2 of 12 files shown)


charlesknipp:
## Notes

@THargreaves I almost managed to get the Levy-SSM working, but fell short when defining the `SSMProblems.logdensity` for the Levy dynamics. I figured since we store the paths, it should be more efficient to grab the previous state's paths. The only issue is when we initialize in the [first draft](https://github.com/TuringLang/AdvancedPS.jl/blob/2880bd3b8bde3cd2739164d5ed181a44032f592e/examples/levy-ssm/script.jl#L164) of this example, we set `path[0] = GammaPath(T[], T[])`, which causes the `meancov` function to error.

I also noticed the docs haven't successfully built in quite a long time, so I update the compat entry for Documenter since that seemed to cause issues with `deploydocs`.

Lastly, feel free to make any changes you see fit.

THargreaves:
> @THargreaves I almost managed to get the Levy-SSM working, but fell short when defining the SSMProblems.logdensity for the Levy dynamics. I figured since we store the paths, it should be more efficient to grab the previous state's paths. The only issue is when we initialize in the [first draft](https://github.com/TuringLang/AdvancedPS.jl/blob/2880bd3b8bde3cd2739164d5ed181a44032f592e/examples/levy-ssm/script.jl#L164) of this example, we set path[0] = GammaPath(T[], T[]), which causes the meancov function to error.

It's interesting coming to review my original code, now that I'm more comfortable with Lévy processes. It actually doesn't look like this implementation is correct.

I wrote the Lévy example before we had a RBPF implemented. The idea was to treat the subordinator jump simulation as part of the randomness of the transition dynamics. This works fine if you are just filtering (only using `simulate`) but the distribution it gives you is actually a conditional one on the specific jumps you generated for that simulation.

This means that the logpdf of the distribution returned by `SSMProblems.distribution` is not the correct logdensity for going from one state to another. In fact, for multivariate linear SDEs, there is no closed form for this logdensity, hence the need for RBPF and storing the jumps explicitly.

I made a proper Lévy SSM example using the RBPF for the Fusion paper. It's currently a bit broken after the interface changes we've made, but I can fix that and link to that repo from SSMProblems.jl README. This can replace this example (especially since Hong suggested we gradually move these Markovian examples out of AdvancedPS.jl).

I suggest given this implementation isn't actually correct, we remove this example.

yebai:
> I suggest given this implementation isn't actually correct, we remove this exampl

Happy to remove the Levy example here and add it to SSMProblems.

penelopeysm (changes requested):
This looks fine to me (at least from a software engineering perspective; I can't comment on actual SSMs). Could you bump the patch version in `Project.toml` please?

PR #125, opened 2026-07-16
Dependency updates and maintenance

shravanngoswamii:
Supersedes #124.

- Bumps the all-julia-packages group, including StatsFuns 2.
- Fixes the docs build: develop the local package before instantiating each example env, so examples resolve against this version's compat rather than the last release.
- Drops Coveralls (Codecov already covers it).
- Drops Requires; the Libtask extension loads through native package extensions.
- Trims compat lower bounds: AbstractMCMC 5, Distributions 0.25, StatsFuns 1.
- Bumps to 0.8.0.


shravanngoswamii:
Well, AdvancedPS is being deprecated so I guess this won't matter anymore. But we can still merge this to give this repo a good updated `bye` in favour of https://github.com/TuringLang/Turing.jl/pull/2848

yebai:
Thanks @shravanngoswamii!

